PR #9, opened 2019-04-12
All recent changes merged

henankf223:
-Add an API to interact with Openfermion
-Add simple input/output helper to help create and print classes
-Add QFT and Reverse QFT for Phase Estimation
-Test cases for api, io and qft
-Add experiment class (python side)
-Add test case for experiment class
-Add trotterization class (python side)
-Add test case for 1st order trotterization 
-Documented


src/qforte/utils/exponentiate.py
```diff
@@ -0,0 +1,74 @@
...
+def exponentiate_single_term(param, term):
+    """
+    A function which returns an exponentiated version of a single (param * term)
+    from a parameterized qubit state preparation circuit.
+
+    :param term: the term to be exponentiated
+    """
+
+    # TODO: look into multiplication by 1.0j? (Nick)
+    # TODO: This code is very similar to that used in PyQuil,
```
nstair: @fevangelista, as noted in the TODO, this chunk of code is similar to PyQuill's, what is the best way to reference it (or should it be rewritten?)

nstair: @fevangelista, what should we do given that this is a similar patch of code to that found in PyQuil? Reference their code? or change ours?

fevangelista: I think this is a standard implementation of the exponential as found in the literature. I have some suggestions for improving it a bit.

src/qforte/quantum_computer.cc
```diff
@@ -79,6 +122,92 @@ void QuantumComputer::apply_gate(const QuantumGate& qg) {
...
+            value *= 1. - 2. * static_cast<double>(basis_[measurement].get_bit(target_qubit));
+        }
+        results[k] = value;
+    }
+
+    coeff_ = old_coeff;
+    return results;
+}
+
+std::vector<double> QuantumComputer::measure_rotated_circuit(const QuantumCircuit& rot_qc, size_t n_measurements) {
```
fevangelista: What does this function do?

src/qforte/quantum_computer.h
```diff
@@ -76,24 +76,49 @@ class QuantumCircuit {
     /// return a vector of gates
     const std::vector<QuantumGate>& gates() const { return gates_; }
 
+    /// return a vector of reversed gates
+    const std::vector<QuantumGate>& reversed_gates() const { return rev_copy_; }
+
+    /// reset the circuit with a new set of parameters
+    void set_parameters(const std::vector<double>& params);
+
+    /// set the reversed gates
```
fevangelista: Why needed? This can be buried in `reversed_gates()` so that is automatically called.

src/qforte/quantum_computer.h
```diff
@@ -76,24 +76,49 @@ class QuantumCircuit {
...
+    /// reversed list of gates
+    std::vector<QuantumGate> rev_copy_;
 };
 
 class QuantumOperator {
   public:
-    /// default constructor: creates an empty quantum operator
-    QuantumOperator() {}
+    /// creates an empty quantum operator
+    QuantumOperator(bool mirror = false) { is_mirror_ = mirror; }
```
fevangelista: Change mirror once new name is identified.

src/qforte/utils/exponentiate.py
```diff
@@ -0,0 +1,74 @@
...
+    # make sure this is ok (Nick)
+
+    if not numpy.isclose(numpy.imag(param), 0.0):
```
fevangelista: Unclear why this is needed.

PR #12, opened 2019-04-23
QFT reverse and test case fix

henankf223:
-Use adjoint function in reverse qft
-Fix the qft test case to include reverse qft

### DIFF
--- a/src/qforte/utils/qft.py
+++ b/src/qforte/utils/qft.py
@@ -31,11 +31,8 @@ def qft_circuit(n, direct):
             qft_circ.add_gate(qforte.make_gate('SWAP', i, n-1-i))
 
     if direct == 'forward':
-        qforte.smart_print(qft_circ)
         return qft_circ
     if direct == 'reverse':
-        qforte.smart_print(qft_circ)
-        qforte.smart_print(qft_circ.adjoint())
         return qft_circ.adjoint()
 
     return qft_circ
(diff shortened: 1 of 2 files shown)


src/qforte/utils/qft.py
```diff
@@ -33,6 +30,14 @@ def qft_circuit(n, direct):
         for i in range(int((n-1)/2)):
             qft_circ.add_gate(qforte.make_gate('SWAP', i, n-1-i))
 
+    if direct == 'forward':
+        qforte.smart_print(qft_circ)
```
fevangelista: We probably want to remove these print statements.

tests/qft_tests.py
```diff
@@ -7,8 +7,13 @@ def test_qft(self):
         trial_circ = build_circuit('X_1')
         trial_state.apply_circuit(trial_circ)
 
+        # verify transformation
         qft(trial_state, 4)
 
+        # test unitarity
+        qft(trial_state, 3)
+        rev_qft(trial_state, 3)
+
```
fevangelista: I would test the qft more extensively. For example, can you do a QFT of a more complicated state and check if the coefficient of the final wave function match some reference values? Same for the inverse. However, I would also keep a test where you start from a state and do qft followed by reverse qft to make sure the final state is the same as the beginning one. We can talk about it if you want.

PR #16, opened 2019-08-02
Faster bit manipulation routines

fevangelista:


src/qforte/quantum_computer.cc
```diff
@@ -132,11 +152,127 @@ void QuantumComputer::apply_1qubit_gate(const QuantumGate& qg) {
             }
         }
     }
+    none_ops_++;
+}
+
+// void QuantumComputer::apply_1qubit_gate_fast(const QuantumGate& qg) {
```
henankf223: Is it a good idea to delete this function? It would be clear if we only retain apply_***_gate for fast operation and apply_***_gate_safe for the older approach.

src/qforte/quantum_computer.h
```diff
@@ -87,14 +93,21 @@ class QuantumComputer {
     size_t none_ops_ = 0;
     /// the number of two-qubit operations
     size_t ntwo_ops_ = 0;
-
     /// the threshold for priting a determinant
     double print_threshold_ = 0.0;
-
+    /// the threshold for doing operations with elemets of gate matricies
```
henankf223: elements?

src/qforte/quantum_computer.h
```diff
@@ -87,14 +93,21 @@ class QuantumComputer {
...
     double compute_threshold_ = 1.0e-16;
 
+    /// apply a 1qubit gate to the quantum computer with standard algorithm
+    void apply_1qubit_gate_safe(const QuantumGate& qg);
+
+    /// apply a 1qubit gate to the quantum computer with optemized algorithm
```
henankf223: optimized

src/qforte/quantum_computer.h
```diff
@@ -22,13 +22,19 @@ class QuantumComputer {
...
+    void apply_circuit_safe(const QuantumCircuit& qc);
+
+    /// apply a quantum circuit to the current state with optimized algorithm
     void apply_circuit(const QuantumCircuit& qc);
 
-    /// apply a gate to the quantum computer
+    /// apply a gate to the quantum computer with standard algorithm
+    void apply_gate_safe(const QuantumGate& qg);
+
+    /// apply a gate to the quantum computer with optemized algorithm
```
henankf223: optimized

src/qforte/quantum_computer.cc
nstair: I will delete it, Idea may me important for future optimizations, but yah it's just straggler code at the moment

src/qforte/quantum_computer.h
nstair: nice catch haha

PR #51, opened 2021-01-13
Fix import issues with latest version of openfermion

fevangelista:
In the latest version of open fermion the path to several functions has changed and these changes are necessary just to import qforte.


JonathonMisiewicz:
I couldn't get qforte to compile without this, either.

Nick, let me know if it's not my place to be merging PRs, but I think this was a simple enough one.

PR #66, opened 2021-03-31
MRSQK Cleanup and Bugfix

JonathonMisiewicz:
This is another case where I was a cleaning up a method, then stumbled cross a bug, so I want to make the PR early. [This code](https://github.com/evangelistalab/qforte/blob/master/src/qforte/qkd/mrsqk.py#L617-L619) performs some normalization operation on `basis_coeff_vec_lst`, which is never used again. I believe this was supposed to be on `basis_coeff_mat`, which is used later. This PR eliminates `basis_coeff_vec_lst` in favor of `basis_coeff_mat`, so that error _cannot_ happen.

Please check that's correct. I'm still trying to figure out the "big picture" of the numerical linear algebra operations here.

### DIFF
--- a/src/qforte/ite/qite.py
+++ b/src/qforte/ite/qite.py
@@ -12,7 +12,35 @@ from scipy.linalg import lstsq
 
 from qforte.maths.eigsolve import canonical_geig_solve
 
+### Throughout this file, we'll refer to DOI 10.1038/s41567-019-0704-4 as Motta.
+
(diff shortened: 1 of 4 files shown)


nstair (commented):
Nice catch for QITE, I think I introduced that bug recently in the qite_speedup PR. I want to look at the MRSQK ref-selection code a bit more and do some testing before merging your changes, maybe chat tomorrow or something.

PR #71, opened 2021-05-05
Cleanup and frozen_virtual example

### LINKED ISSUE #69: Clean `system_factory.py`
A very nice cleanup that I'll do after the qforte issue I'm currently working on cleaning:

First, instead of passing arguments to the molecule adapters explicitly, [system_factory](https://github.com/evangelistalab/qforte/blob/master/src/qforte/system/system_factory.py) should send all of them kwargs. Now, caller-side, all three adapters take the same signature.
Second, replace the if-then structure with a dictionary that maps strings to the adapter to call. Once you have the string, read the adapter from the dictionary.

This is much less code and has much less duplication. As a further advantage, the list of allowed adapter types, which we need for the error message, can be read from the dictionary. Less code duplication!

JonathonMisiewicz:
This PR cleans up some of the adapter code and adds a test of using the frozen_virtual approximation. It took me too long to figure out how it worked. Closes #69.

### DIFF
--- a/src/qforte/adapters/molecule_adapters.py
+++ b/src/qforte/adapters/molecule_adapters.py
@@ -41,7 +41,7 @@ except:
 class MolAdapter(ABC):
     """Abstract class for the aquiring system information from external electronic
     structure calculations. The run() method calculates the desired properties and
-    data for the molecular system. Infromation is then stored in the
+    data for the molecular system. Information is then stored in the
     self._qforte_mol attribuite.
     """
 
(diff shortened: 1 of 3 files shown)


src/qforte/adapters/molecule_adapters.py
```diff
@@ -343,14 +326,9 @@ def run(self, **kwargs):
...
         # Make hf_reference
-        hf_reference = []
-        for a in range(2*nmo):
-            if(a<nel):
-                hf_reference.append(1)
-            else:
-                hf_reference.append(0)
-
-        # Build second quantized Hamiltonain
+        hf_reference = [1] * nel + [0] * (2 * nmo - nel)
```
nstair, line 329: This is cool, have never initialized a list like this before!

PR #73, opened 2021-05-13
Remove references to unused mo_oei and mo_tei member variables

JonathonMisiewicz:
Exactly what it says in the title. Nick, was there some reason why these variables were added in the first place?


nstair (approved):
I thought they would potentially be useful to have access to for some applications in the future. We can add these attributes later if it comes up.

PR #74, opened 2021-05-17
Simplified Molecule Creation

### LINKED ISSUE #70: Condense MolAdapter Init and Run
The [various `MolAdapter` subclasses](https://github.com/evangelistalab/qforte/blob/master/src/qforte/adapters/molecule_adapters.py) need to be initialized and then run. Both the initializer and runner take keyword arguments, but it isn't obvious to me when an argument should go into `__init__` and when it should go into `run`. Further, I can't imagine a use case for calling `run` multiple times, or _not_ immediately after `__init__`. To me, these indicate that the `run` functionality needs to be combined into `__init__`.

@nstair, thoughts?

JonathonMisiewicz:
This PR closes #70 and is API-breaking.

In the old API, `system_factory` would return a `molecule_adapter` object, the object's `run` method would be called, and then its `get_molecule` method would be called. The `molecule_adapter` object is good for absolutely nothing else. This PR streamlines the procedure: The various `molecule_adapter` objects are now just functions that return a molecule. The associated class hierarchy is gone, and this three-step procedure to create a molecule is reduced to just one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,9 +54,7 @@ from qforte.system import system_factory
 H2geom = [('H', (0., 0., 0.)), ('H', (0., 0., 1.50))]
 H2ref = [1,1,0,0]
 
-adapter = system_factory(mol_geometry=H2geom)
-adapter.run()
-H2mol = adapter.get_molecule()
+H2mol = system_factory(mol_geometry=H2geom)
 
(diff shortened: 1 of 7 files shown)


nstair (commented):
Looks good

PR #77, opened 2021-05-18
Psi4 is now symmetry-compatible

### LINKED ISSUE #76: Molecule Creation From Psi Not Symmetry-Tolerant
It's currently not possible to use a Psi4 molecule with symmetry.

When you try, for example,
```
    mol = qforte.system.system_factory(system_type = 'molecule',
                                 build_type = 'psi4',
                                 basis='cc-pvdz',
                                 mol_geometry = [('He', (0, 0, 0))],
                                 symmetry = "c2v")
```

you hit

```
Traceback (most recent call last):
  File "/Users/jonathonmisiewicz/qforte/tests/uccsd_test.py", line 113, in <module>
    mol = system_factory(system_type = 'molecule',
  File "/Users/jonathonmisiewicz/qforte/src/qforte/system/system_factory.py", line 46, in system_factory
    return adapter(**kwargs)
  File "/Users/jonathonmisiewicz/qforte/src/qforte/adapters/molecule_adapters.py", line 293, in create_psi_mol
    mo_oeis = np.einsum('uj,vi,uv', C, C, mo_oeis)
  File "<__array_function__ internals>", line 5, in einsum
  File "/Users/jonathonmisiewicz/miniconda3/envs/p4dev/lib/python3.9/site-packages/numpy/core/einsumfunc.py", line 1350, in einsum
    return c_einsum(*operands, **kwargs)
ValueError: einstein sum subscripts string contains too many subscripts for operand 0
```

There may be other points where using symmetry causes computation failure.

JonathonMisiewicz:
Closes #76 and knocks off an item on @fevangelista's wishlist: it's now possible to tell Psi4 to use symmetry for its own computations! QForte doesn't take advantage of this at present, but at least now it _can_ use orbitals from a computation with symmetry.

This also amends the recent Psi4 molecule test case to employ symmetry.

### DIFF
--- a/src/qforte/adapters/molecule_adapters.py
+++ b/src/qforte/adapters/molecule_adapters.py
@@ -279,7 +279,7 @@ def create_psi_mol(**kwargs):
     # Get integrals using MintsHelper.
     mints = psi4.core.MintsHelper(p4_wfn.basisset())
 
-    C = p4_wfn.Ca()
+    C = p4_wfn.Ca_subset("AO", "ALL")
 
     scalars = p4_wfn.scalar_variables()
 
--- a/tests/uccsd_test.py
+++ b/tests/uccsd_test.py
@@ -44,7 +44,8 @@ class UccTests(unittest.TestCase):
                                      build_type = 'psi4',
                                      basis='cc-pvdz',
                                      mol_geometry = [('He', (0, 0, 0))],
-                                     filename=data_path)
+                                     filename=data_path,
+                                     symmetry = "c2v")
 
         ref = [1,1,0,0,0,0,0,0,0,0]
 


fevangelista:
The looks nice. But are all the changes committed? It's strange that it would take only one line to enable symmetry in psi4.

JonathonMisiewicz:
I oversimplified earlier: we could already tell Psi4 to run in symmetry. [This line is how the symmetry keyword gets set](https://github.com/evangelistalab/qforte/blob/master/src/qforte/adapters/molecule_adapters.py#L216) and [this line is how it gets sent to Psi4](https://github.com/evangelistalab/qforte/blob/master/src/qforte/adapters/molecule_adapters.py#L246). So we had the C matrix with symmetry already.

What is new is being able to use the C matrix to transform the integrals without crashing, and that's all my line of code is needed for.

fevangelista (approved):

fevangelista:
Great!

PR #80, opened 2021-05-21
Remove old, buggy code

JonathonMisiewicz:
ADAPT-VQE used to support different operator selection types. Only one of them is actually working, so this PR majorly simplifies the code to only have the one that works. I want approval from Nick on this, in case he thinks the code is salvageable.


JonathonMisiewicz:
For the record, there's more UCC cleanup to do. This is just worth its own PR because it technically is feature removal.

PR #81, opened 2021-05-21
Removes need for user defined reference list.

nstair:
This PR removes the need for the user to specify and pass a reference list (list of 1's and 0's that usually defines the hartree fock state), it is minorly API breaking. The reference list is now automatically generated in the functions that build the molecule objects, but it is still possible to use a specified reference list by passing one. Test cases have been updated to reflect the new API.

### DIFF
--- a/src/qforte/abc/algorithm.py
+++ b/src/qforte/abc/algorithm.py
@@ -51,7 +51,7 @@ class Algorithm(ABC):
 
     def __init__(self,
                  system,
-                 reference,
+                 reference=None,
                  trial_state_type='reference',
                  trotter_order=1,
                  trotter_number=1,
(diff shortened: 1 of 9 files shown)


src/qforte/abc/algorithm.py
```diff
@@ -60,10 +60,14 @@ def __init__(self,
                  print_summary_file=False):
 
         self._sys = system
-        self._ref = reference
-        self._nqb = len(reference)
+        if(reference==None):
+            self._ref = system.get_hf_reference()
+        else:
+            self._ref = reference
```
JonathonMisiewicz, line 66:
```suggestion
        self._ref = system.get_hf_reference() if reference is None else reference
```

JonathonMisiewicz (approved):
I have one suggestion, but LGTM otherwise.

JonathonMisiewicz:
@nstair For future reference, when I make a suggestion like that, there should be a button to commit the suggestion. That will automatically add a commit making the change to your remote fork.

PR #82, opened 2021-05-22
Cleanup UCC

JonathonMisiewicz:
This cleans up anything and everything involving UCC to eliminate code duplication. We now have an `algorithm` subclass called `AnsatzAlgorithm` for any algorithm that requires an ansatz. We have a further subclass for this, UCC, for functions further common to when that ansatz is UCC.

Hooray for -400 LoC. Tests still pass.


nstair (approved):
Looks great, I like the AnsatzAlgorithm class idea. This PR feels like finally closing all of those unused browser tabs I have open.

PR #83, opened 2021-05-24
Rename make_gate and make input more friendly

fevangelista:
This PR renames `qforte.make_gate() -> qforte.gate()` and allows `gate` to handle the following cases:

- `gate(type, control)`: only for single qubit gates
- `gate(type, control, target)`: only for two-qubit gates that require no parameters
- `gate(type, control, parameter)`: only for single qubit gates
- `gate(type, control, target,parameter)`: for any type of gate (there is an extra layer of checks at the C++ side)

- [x] All tests pass.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,8 +44,8 @@ import qforte
 
 # Construct a Bell state.
 computer = qforte.QuantumComputer(2)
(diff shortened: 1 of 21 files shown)


nstair (approved):
This looks good to me.

PR #86, opened 2021-05-25
Make tests findable

JonathonMisiewicz:
## Description
`pytest` can now pick up all of the tests.

I had to rename one function to avoid `pytest` thinking that was a test. `comprehensive_gates_test.py` is too copy-paste as-is, but we can make an issue for that. I have other PRs to work on now.

## Checklist
- [x] Ready to go!

The rest of the checklist isn't really applicable.


fevangelista:
Looks good to me.

PR #90, opened 2021-05-26
Remove Particle-Conserving Assumption

JonathonMisiewicz:
## Description
The primary purpose of this PR is to remove from the SQOperator class (and callers) the assumption that the operators are particle-conserving.

Secondary features of interest: QITE no longer accepts a SQ Hamiltonian argument (which it didn't even use). The PR template is modified from the Forte template to be more relevant for QForte. You can't document new features in the manual if there isn't a manual.

## User Notes
- [x] New Feature: SQ Operators need not be particle conserving
- [x] Major API change! All SQ Operator constructors now require separate lists of creation indices and annihilation indices
- [x] Passing a SQ Hamiltonian to QITE is no longer allowed. This had no effect in previous code.

## Checklist
- [x] Documented source code
- [x] Checked for redundant headers and imports
- [x] Checked for consistency in the formatting of the output file
- [x] Ready to go!

### DIFF
--- a/src/qforte/sq_operator.h
+++ b/src/qforte/sq_operator.h
@@ -19,12 +19,8 @@ class SQOperator {
     /// default constructor: creates an empty second quantized operator
     SQOperator() {}
 
-    /// TODO: implement
-    /// build from a string of open fermion qubit operators
-    // void build_from_openferm_str(std::string op) {}
-
-    /// add one product of anihilators and/or creators to the second quantized operator
-    void add_term(std::complex<double> coeff, const std::vector<size_t>& ac_ops);
+    /// add one product of annihilators and/or creators to the second quantized operator
+    void add_term(std::complex<double> coeff, const std::vector<size_t>& cre_ops, const std::vector<size_t>& ann_ops);
 
     /// add an second quantized operator to the second quantized operator
     void add_op(const SQOperator& sqo);
@@ -36,11 +32,11 @@ class SQOperator {
     void mult_coeffs(const std::complex<double>& multiplier);
 
     /// return a vector of terms and their coefficients
-    const std::vector<std::pair< std::complex<double>, std::vector<size_t>>>& terms() const;
+    const std::vector<std::tuple< std::complex<double>, std::vector<size_t>, std::vector<size_t>>>& terms() const;
 
     /// Put a single term into "canonical" form. Canonical form orders orbital indices
     /// descending.
-    void canonical_order_single_term(std::pair< std::complex<double>, std::vector<size_t>>& term );
+    void canonical_order_single_term(std::tuple< std::complex<double>, std::vector<size_t>, std::vector<size_t>>& term );
 
     /// Canonicalize each term. The order of the terms is unaffected.
     void canonical_order();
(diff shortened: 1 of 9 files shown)


src/qforte/sq_operator.h
```diff
@@ -19,12 +19,8 @@ class SQOperator {
     /// default constructor: creates an empty second quantized operator
     SQOperator() {}
 
-    /// TODO: implement
-    /// build from a string of open fermion qubit operators
-    // void build_from_openferm_str(std::string op) {}
-
-    /// add one product of anihilators and/or creators to the second quantized operator
-    void add_term(std::complex<double> coeff, const std::vector<size_t>& ac_ops);
+    /// add one product of annihilators and/or creators to the second quantized operator
```
fevangelista, line 22: Can we please add some text explaining on how the operators are ordered? For example, in Forte we have something like:
```
 * A data structure used to represent a second quantized operator string like
 *
 * ... op(2) op(1) op(0), where op(i) = a^\dagger_(orb_i, spin_i) or a_(orb_i, spin_i)
 * ```

Comment (fevangelista) [reply src/qforte/sq_operator.h:22]: To clarify, this should probably appear at the beginning of the class.

File: src/qforte/sq_operator.h [hunk 2]
```diff
@@ -19,12 +19,8 @@ class SQOperator {
...
+    void add_term(std::complex<double> coeff, const std::vector<size_t>& cre_ops, const std::vector<size_t>& ann_ops);
 
     /// add an second quantized operator to the second quantized operator
```
fevangelista, line 25:
```suggestion
    /// add an second quantized operator to this second quantized operator
```

src/qforte/sq_operator.h
```diff
@@ -59,10 +55,16 @@ class SQOperator {
     /// The linear combination of second quantized operators. Stored in pairs of
```
fevangelista, line 55:
```suggestion
    /// The linear combination of second quantized operators. Stored as a tuple of
```

src/qforte/sq_operator.h
```diff
@@ -59,10 +55,16 @@ class SQOperator {
...
     /// coefficients, and then a vector of N created indices, followed by N annihilated indices.
     /// Orbital indices start at zero.
-    std::vector<std::pair< std::complex<double>, std::vector<size_t>>> terms_;
+    std::vector<std::tuple< std::complex<double>, std::vector<size_t>, std::vector<size_t>>> terms_;
 
     /// Calculate the parity of permutation p
-    bool permutive_sign_change(std::vector<int> p);
+    bool permutive_sign_change(std::vector<int> p) const;
```
fevangelista, line 61: This does not really belong here since it is a general function. I would suggest to move to a `combinatorial.cc` or `math.cc` file. This can be done in another PR.
```suggestion
    bool permutation_sign(std::vector<int> p) const;
```

fevangelista (changes requested):
Thank you for working on this and getting it done so quickly. I left a few minor comments. The new design looks great and I think we could easily port it to the bit-storage solution later on.

src/qforte/sq_operator.h
JonathonMisiewicz: Yeah, let's defer moving this to another PR, when we have a better idea what this other function might look like. Done the renaming, though.

nstair (approved):
Looks good to me, thanks for the big PR!

fevangelista (approved):

PR #91, opened 2021-05-26
Rename QuantumClasses

nstair:
## Description
This PR renames several of the core classes in QForte. 

Namely:
QuantumBasis->Basis
QuantumGate->Gate
QuantumCircuit->Circuit
QuantumOperator->QubitOperator
QuantumOpPool->QubitOpPool

## User Notes
- [ ] Features added
- [x] Changes to compilation (if any)

## Checklist
- [ ] Documented source code
- [ ] Ready to go!


JonathonMisiewicz:
I won't bother checking this if tests run, and this is the simple search-replace job I think it is.

The one thing I'm not sure about is renaming `QubitBasis` to `Basis`. `TensorBasis` sounds more appropriate to me, in case we need to add other basis states, like if we rename `SqOperator` to `SqOperatorBasis`.

fevangelista:
How about `QubitBasis` (Before we had `QuantumBasis`)?

JonathonMisiewicz:
`QubitBasis` is fine right now. The class is abstract enough that it may work as a basis of determinants when you get the fermionic quantum computer running. If we go that route, we'll need to discuss names again, but I can accept leaving that to the future and going with `QubitBasis` for the immediate future.

PR #110, opened 2021-08-05
One hopefully last bit of pre-DIIS refactoring

JonathonMisiewicz:
## Description
This PR moves removes some code duplication and also moves the ability to take MP denominators off of PQE and onto the general AnsatzAlgorithm, right by the ability to compute orbital energies. I also did some minor code cleanup in `uccnvqe.py`.

I expect that the next PR will be the one that lets VQE optimize with DIIS. This will be the largest of the PRs I've submitted in the last day or two, but I don't see a good way to break it down any further.

## User Notes
- [x] Generalized `get_res_over_mpdenom` out of PQE 

## Checklist
- [x] Ready to go!

### DIFF
--- a/src/qforte/abc/ansatz.py
+++ b/src/qforte/abc/ansatz.py
@@ -62,3 +62,37 @@ class UCC:
 
             print(f'  {i:3}     {ei:+16.12f}')
             self._orb_e.append(ei)
+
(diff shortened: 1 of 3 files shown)


imagoulas (approved):
Approved and merged.

PR #112, opened 2021-08-07
More Variable Synchronizing

JonathonMisiewicz:
## Description
Three more sets of renamings to synchronize variables across PQE and VQE, which will hopefully be the last bit of cleanup before the work I actually care about.

The review for this can absolutely wait until Monday, but now I won't have this nagging at me all weekend.

## Checklist
- [x] Ready to go!

### DIFF
--- a/src/qforte/abc/algorithm.py
+++ b/src/qforte/abc/algorithm.py
@@ -204,6 +204,11 @@ class AnsatzAlgorithm(Algorithm):
     _opt_maxiter : int
         The maximum number of iterations for the classical optimizer.
 
+    _opt_thresh : float
+        The numerical convergence threshold for the specified classical
+        optimization algorithm. Is usually the norm of the gradient, but
+        is algorithm dependant, see scipy.minimize.optimize for detials.
+
     _pool : list of tuple(complex, SqOperator)
         The linear combination of (optionally symmetrized) single and double
         excitation operators to consider. This is represented as a list.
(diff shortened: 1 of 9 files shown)


src/qforte/abc/algorithm.py
```diff
@@ -56,6 +56,11 @@ class Algorithm(ABC):
         The number of pauli terms (Hermitian products of Pauli X, Y, and/or Z gates)
         mesaured over the entire algorithm.
 
+    _res_vec_evals : int
+        The total number of times the entire residual was evaluated.
+
+    _res_m_evals : int
+        The total number of times an individal residual element was evaluated.
```
imagoulas, line 63: "individal" -> "individual"

src/qforte/abc/algorithm.py
```diff
@@ -199,6 +204,11 @@ class AnsatzAlgorithm(Algorithm):
     _opt_maxiter : int
         The maximum number of iterations for the classical optimizer.
 
+    _opt_thresh : float
+        The numerical convergence threshold for the specified classical
+        optimization algorithm. Is usually the norm of the gradient, but
+        is algorithm dependant, see scipy.minimize.optimize for detials.
```
imagoulas, line 210: "detials" => "details"

imagoulas (approved):
Everything looks good to me. I added a couple of comments regarding typos in the comments to the source code (I noticed that in earlier PRs you were also fixing those).

JonathonMisiewicz:
Thanks for the catches, Ilias. Nick left a lot of them (that's what happens when you're coding alone), so every one we can catch helps.

PR #113, opened 2021-08-10
DIIS for VQE

### LINKED ISSUE #105: DIIS for VQE
Because we have DIIS-based algorithm for PQE, let's get one for VQE. I'm paranoid about whether the reason for the success of PQE over VQE in [Section IIIA](https://journals.aps.org/prxquantum/abstract/10.1103/PRXQuantum.2.030301) is due to the residual conditions or the algorithm used to achieve them.

JonathonMisiewicz:
## Description
Closes #105: we can now use DIIS on VQE!

## User Notes
- [x] `diis_solve` algorithm for VQE

## Checklist
- [x] Ready to go!

### DIFF
--- a/src/qforte/abc/vqeabc.py
+++ b/src/qforte/abc/vqeabc.py
@@ -63,9 +63,6 @@ class VQE(AnsatzAlgorithm):
         if self._converged is None:
             raise NotImplementedError('Concrete VQE class must define self._converged attribute.')
 
-        if self._final_result is None:
-            raise NotImplementedError('Concrete VQE class must define self._final_result attribute.')
-
         if self._opt_maxiter is None:
             raise NotImplementedError('Concrete VQE class must define self._opt_maxiter attribute.')
 
--- a/src/qforte/ucc/uccnpqe.py
+++ b/src/qforte/ucc/uccnpqe.py
@@ -9,6 +9,7 @@ import qforte
 from qforte.abc.uccpqeabc import UCCPQE
 
 from qforte.experiment import *
+from qforte.maths import optimizer
 from qforte.utils.transforms import *
 from qforte.utils.op_pools import *
 from qforte.utils.state_prep import *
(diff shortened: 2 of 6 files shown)


src/qforte/ucc/uccnvqe.py
```diff
@@ -237,8 +244,12 @@ def get_num_ham_measurements(self):
         """Returns the total number of times the energy was evaluated via
         measurement of the Hamiltoanin.
```
imagoulas, line 245: Typo:
Hamiltoanin -> Hamiltonian

src/qforte/ucc/uccnvqe.py
```diff
@@ -248,3 +259,5 @@ def get_num_commut_measurements(self):
         # else:
         #     return 0
         return 0
+
+UCCNVQE.diis_solve = optimizer.diis_solve
```
imagoulas, line 263: See similar comment for uccnpqe.py

tests/uccsd_test.py
```diff
@@ -32,6 +32,28 @@ def test_He_uccsd_vqe_exact(self):
         Egs = Egs_elec + Enuc
         self.assertLess(abs(Egs-Efci), 1.0e-11)
 
+    def test_He_uccsd_vqe_exact(self):
+        print('\n')
+        # The FCI energy for He atom in a cc-pvdz basis
+        Efci = -2.887594831090938
+        # The Nuclear repulsion energy
+        Enuc =  0.0
```
imagoulas, line 40: Line 40 could be removed and the following additional modifications should be introduced:
line 53: Egs_elec -> Egs
line 54: remove

The same modifications could be made to the following test as well, for consistency. On a side note, I believe that line 68 should be removed; since in this test we are using psi4 to generate the one- and two-electron integrals, we should not be reading from a file.

imagoulas (commented):
I made a few comments, most about typos and a couple about the code itself. I also made a few comments with questions aimed at educating myself.

In the spqe.py file there is a diis_solve function defined that is very similar, if not identical, to the one that was originally in uccnpqe. Eventually, this function could be removed and call the one that you created in optimizer.py.

Final remark: does the __init__.py file in the maths folder need to be modified since you added optimizer.py? The answer might be obvious, but I am not yet familiar with these files.

src/qforte/ucc/uccnpqe.py
```diff
@@ -346,3 +265,5 @@ def initialize_ansatz(self):
         for l in range(len(self._pool)):
             self._tops.append(l)
             self._tamps.append(0.0)
+
+UCCNPQE.diis_solve = optimizer.diis_solve
```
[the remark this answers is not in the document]
JonathonMisiewicz: This line is needed precisely because UCCNPQE does not have a `diis_solve` attribute. This line _creates_ the attribute by setting it to be `optimizer.diis_solve`.

src/qforte/abc/vqeabc.py
```diff
@@ -63,9 +63,6 @@ def verify_required_VQE_attributes(self):
         if self._converged is None:
             raise NotImplementedError('Concrete VQE class must define self._converged attribute.')
 
-        if self._final_result is None:
-            raise NotImplementedError('Concrete VQE class must define self._final_result attribute.')
```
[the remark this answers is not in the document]
JonathonMisiewicz: No. The `_final_result` object is created by SciPy when you use it as a minimizer. When you use a non-SciPy minimizer, no `_final_result` is created.

JonathonMisiewicz:
All comments addressed.

imagoulas (approved):
Thank you! Approved and merged!

Issue #119, opened 2021-08-18
Use Properties in Molecule


JonathonMisiewicz:
`molecular_info.py` defines many get and set methods. These should be `@property`.

Issue #118, opened 2021-08-18
Build More Documentation


JonathonMisiewicz:
For some reason, the built docs are incomplete and don't include, for example, the Molecule class.

Issue #116, opened 2021-08-18
Create ABC for System


JonathonMisiewicz:
While there is a molecule class, several QForte algorithms are designed to work for more general _systems_. We can't guarantee this, because there isn't a general system class.

I need to create a class for system and confirm that everything that is supposed to work for a system, not a molecule in particular, actually does.

PR #120, opened 2021-08-18
Add System class

JonathonMisiewicz:
## Description
This PR adds a new class to QForte for generic systems. The Molecule now inherits from this. This PR is a step towards #116, but does not confirm that Forte's algorithms are valid on a general system.

Documentation is inadequate at present. My ability to check documentation is blocked by #118. This PR blocks #119.

Sorry Ilias, but this is going to kick off another long string of QForte PRs from me.

## User Notes
- [x] Added the System class

## Checklist
- [x] Documented source code
- [x] Ready to go!

### DIFF
--- a/src/qforte/system/molecular_info.py
+++ b/src/qforte/system/molecular_info.py
@@ -1,12 +1,38 @@
 """
-A class for the system information, either a molecule or a lattice system such as
-Hubbard model.
+Classes for system information, e.g., molecule or Hubbard model.
 """
 
 # TODO: Documentation needs to be fixed, attributes should be listed below
 #       as opposed to arguments for __init__() (Nick).
 
-class Molecule(object):
+class System(object):
+    """Class for a generic quantum many-body system."""
+
+    def set_fci_energy(self, fci_energy):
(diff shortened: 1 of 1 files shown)


src/qforte/system/molecular_info.py
```diff
@@ -120,19 +134,10 @@ def set_cisd_energy(self, cisd_energy):
...
         return self._ccsd_singles, self._ccsd_doubles
 
-    def get_hamiltonian(self):
-        return self._hamiltonian
-
-    def get_sq_hamiltonian(self):
-        return self._sq_hamiltonian
-
     def get_sq_of_ham(self):
         return self._sq_of_ham
```
imagoulas, line 142: Should this function be included in the System class and removed from the Molecule class, as was the case with set_sq_of_ham?

imagoulas (approved):
Everything looks good, but I did leave a minor comment in the code,

src/qforte/system/molecular_info.py
JonathonMisiewicz: Good call. Will fix that, then merge this in.

PR #121, opened 2021-08-18
Remove `_nel`

JonathonMisiewicz:
## Description
This PR removes the `set_nel` function, which set `_nel`. We can re-add it if we need it. Right now, the attribute is used literally nowhere in the code.

## User Notes
- [x] Removed `set_nel`

## Checklist
- [x] Ready to go!


imagoulas (approved):
Approved!

PR #131, opened 2021-08-27
Support custom operator pools

JonathonMisiewicz:
## Description
This PR adds support for users passing their own operator pools. I'll want to allow QubitOpPools soon, but this is a start.

## User Notes
- [x] Users can add their own operator pools! 

## Checklist
- [x] Ready to go!

### DIFF
--- a/src/qforte/abc/algorithm.py
+++ b/src/qforte/abc/algorithm.py
@@ -214,7 +214,7 @@ class AnsatzAlgorithm(Algorithm):
         excitation operators to consider. This is represented as a list.
         Each entry is a pair of a complex coefficient and an SqOperator object.
 
-    _pool_obj : SqOpPool
+    _pool_obj : SQOpPool
         A pool of second quantized operators we use in the ansatz.
 
     _tops : list
@@ -249,11 +249,12 @@ class AnsatzAlgorithm(Algorithm):
         """ This function populates an operator pool with SQOperator objects.
         """
 
-        self._pool_obj = qf.SQOpPool()
-        self._pool_obj.set_orb_spaces(self._ref)
-
         if self._pool_type in {'sa_SD', 'GSD', 'SD', 'SDT', 'SDTQ', 'SDTQP', 'SDTQPH'}:
+            self._pool_obj = qf.SQOpPool()
+            self._pool_obj.set_orb_spaces(self._ref)
             self._pool_obj.fill_pool(self._pool_type)
+        elif isinstance(self._pool_type, qf.SQOpPool):
+            self._pool_obj = self._pool_type
         else:
             raise ValueError('Invalid operator pool type specified.')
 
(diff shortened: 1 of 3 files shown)


imagoulas (commented):
The code looks good to me. Nevertheless, I run into a few unexpected issues when running the tests.

1) I got an error saying that the STO-6G basis set was not found.

I believe that this is caused by the OpenFermion dependency, which forces us (me?) to use python 3.7. As a result, I had to install a somewhat older version of Psi4 that does not include the sto-6g.gbs file, which was added to Psi4 on March 2020 (PR #1853). Mind you that when installing Psi4 I just followed the installation instructions provided by QForte.

Once I manually added the sto-6g.gbs file the test passed successfully. I would suggest replacing STO-6G by STO-3G to avoid this problem. I guess a user can manually install the latest version of Psi4, but our current installation instructions do not suggest so.

2) I have no clue why this is happening, but, in this PR, I get a couple of:
`PSIO_ERROR: 18 (Incorrect block end address)`
I did not have such issues before. The error messages appear in the tests:
test_He_uccsd_vqe_exact_psi
test_srqk_with_uccsdpqe_ref
I have no idea what I did to upset the Psi4 Gods, but apparently they are angry at me! If you have some time before or after the group meeting, We can have a Zoom meeting to show you exactly what I am doing and what the error messages look like.

3) In PR #123, in the fast_qk_test.py test you changed "s=6" to "s=7". I just noticed that, at least on my machine, this test fails:
`AssertionError: 0.00012960177460241695 not less than 0.0001`
If I change back "s=7" to "s=6", the test passes successfully. Can you confirm this issue at your end?

JonathonMisiewicz:
Let's focus on the STO-6G basis set issue first.

_I believe that this is caused by the OpenFermion dependency, which forces us (me?) to use python 3.7. As a result, I had to install a somewhat older version of Psi4 that does not include the sto-6g.gbs file_

The current master release of Psi4 supports Python 3.7, so I don't understand why you're having this issue.

imagoulas:
When I do
`conda list psi4`
I see that the installed version is 1.3.2+ecbda83. I do not know if that helps.
After I got the error about the missing basis set, I checked the folder
`anaconda3/envs/qforte_env/share/psi4/basis`
and so that the sto-6g.gbs file was missing, so I added it manually just to see whether the test would work, and it did. As I said before, I just followed the instructions provided by QForte when installing QForte, OpenFermion, and Psi4. Although the STO-6G basis was incorporated in Psi4 in 2020, I installed QForte in 2021.

JonathonMisiewicz:
I cannot reproduce your reported test fails. Can you give me the _full_ output from the failing tests?

I have the strong suspicion that we need to update the Qforte install instructions.

JonathonMisiewicz:
Reported issue 3 is resolved. I'll deal with 1 and 2 tomorrow. I suspect that we need to update the build instructions to get a more recent version of Psi.

JonathonMisiewicz:
What are the objections to merging at this point? If you check out the latest version of this PR (which includes my last three), all the test failures should be resolved.

imagoulas (approved):
All issues were resolved by PRs #132, #133, and #134. What do you think about adding a description for 'GSD' (see my comment in the uccvqeabc.py file). I leave that up to you.

PR #133, opened 2021-09-02
Fix PSIO Error

JonathonMisiewicz:
## Description
Prevents temporary files from one Psi computation from interfering with the files from another computation. On my computer, this is enough to prevent the bug reported by Ilias, where there is a PSIO error involving `incorrect block end address`.

## Checklist
- [x] Ready to go!


imagoulas (approved):
All tests passed successfully using `python setup.py test` (well, except the "s=7" one, but that was anticipated). Approved and merged.

PR #135, opened 2021-09-03
Added printing statements in tests for debugging

imagoulas:
## Description
I added print statements in the tests files. Now, once each test begins, it will first print the name of the test. This will be useful for debugging purposes when running `python setup.py test`.

I also renamed a couple of tests that had the same name. The new names specify the particular method that is being tested.
 
All tests passed on my machine.

## Checklist
- [ x] Added/updated tests of new features
- [ x] Ready to go!

### DIFF
--- a/tests/adapt_vqe_test.py
+++ b/tests/adapt_vqe_test.py
@@ -11,7 +11,7 @@ data_path = os.path.join(THIS_DIR, 'H4-sto6g-075a.json')
 
 class ADAPTVQETests(unittest.TestCase):
     def test_H4_adapt_vqe_exact(self):
-        print('\n')
+        print('\nSTART test_H4_adapt_vqe_exact\n')
 
         # The FCI energy for H4 at 0.75 Angstrom in a sto-6g basis
         Efci = -2.1628978832666865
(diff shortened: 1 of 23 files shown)


JonathonMisiewicz:
Can you remind me where these test outputs are collected, that you want these print statements to appear?

imagoulas:
When I run `python setup.py test > outputs` the outputs of all tests are directed to outputs. However, it is difficult to figure out which output corresponds to which test. With this PR, one simply needs to search for START and they will immediately find the starting point of each test, accompanied by the name of the function that is responsible for the test.

JonathonMisiewicz (changes requested):
This PR addresses two separate issues. One is that some tests have the same name. I completely support renaming these.

The other issue is that when running `python setup.py test`, when you pipe the output to a file, the test names do not get piped. This PR works around that by _printing_ the test name in the output. However, `python setup.py test` [is deprecated](https://github.com/pypa/setuptools/issues/1684) and will not be supported forever. Instead, users should run `pytest` in the `tests/` directory, and the README should be updated so that we're no longer encouraging users to use deprecated behavior. With this change, do we still need to print out test names?

PR #136, opened 2021-09-09
Added point group symmetry in SQOpPool

imagoulas:
## Description
Added point group symmetry in SQOpPool. Currently, this is only supported for `build_type = 'psi4'`. OpenFermion only works with C1 symmetry and this is yet another reason to stick with pure Psi4.

To make this work, in addition to the new module called `point_groups`, I created three new attributes for the Molecule class, called `point_group`, `orb_irreps`, and `orb_irreps_to_int`. The values of these attributes are obtained by parsing the Psi4 output (I could not find another way to do it; I am open to suggestions). In doing so, I made a modification so now the user can specify the name of the output file, even when they are using `build_type = 'psi4'`.

I also added an additional attribute to the Algorithm class, called `_irrep`. With this attribute, a user can specify the symmetry of the operators in SQOpPool. Currently this only makes sense for `_irrep = 0`, i.e., totally symmetric operators. To make this work in the future, suitable modifications will need to be introduced in the initial guess of the amplitudes so that the HF Slater determinant is "projected out"; starting from 0 amplitudes causes the code to always collapse to HF. As an experiment, I chose a stupid initial guess (setting all amplitudes to 0 except the first one which is set to 1) and it did work for some irreps, but not always. In the future, we should allow the user the option to provide their own initial guess (this will also be helpful for targeting particular excited states).

On a side note, in uccnvqe.py I commented out two lines of code because they were giving me an error when `verbose = True`. It seems that `_pool_obj` has no `print_pool()` attribute defined.

All tests passed successfully on my computer!

## User Notes
- [x ] Features added

## Checklist
- [ ?] Ready to go!


src/qforte/utils/point_groups.py
```diff
@@ -0,0 +1,275 @@
...
+
+    for irrep1 in irrep_to_int.items():
+        print(irrep1[0], '  ', end = '')
+        for irrep2 in irrep_to_int.items():
+            print(int_to_irrep[irrep1[1] ^ irrep2[1]], ' ', end = '')
+        print()
+
+    return None
+
+def symmetry_check(orb_irreps_to_int, annihilate, create):
```
JonathonMisiewicz, line 176: Please add a docstring saying what this function checks. I think you're returning the symmetry of the excitation? If so, please rename this function. "check" is vague.

src/qforte/utils/point_groups.py
```diff
@@ -0,0 +1,275 @@
...
+
+    print('\n')
+
+    for irrep1 in irrep_to_int.items():
+        print(irrep1[0], '  ', end = '')
+        for irrep2 in irrep_to_int.items():
+            print(int_to_irrep[irrep1[1] ^ irrep2[1]], ' ', end = '')
+        print()
+
+    return None
```
JonathonMisiewicz, line 174:
```suggestion
```

JonathonMisiewicz (commented):
I'll have more to say so when I understand how you want this PR to interact with OepnFermion.

src/qforte/abc/algorithm.py
```diff
@@ -258,6 +260,18 @@ def fill_pool(self):
...
+        # If possible, impose symmetry restriction to operator pool
+        try:
+            temp_sq_pool = qf.SQOpPool()
+            for sq_operator in self._pool_obj.terms():
+                create = sq_operator[1].terms()[0][1]
+                annihilate = sq_operator[1].terms()[0][2]
+                if symmetry_check(self._sys.orb_irreps_to_int, create, annihilate) == self._irrep:
+                    temp_sq_pool.add(sq_operator[0], sq_operator[1])
+            self._pool_obj = temp_sq_pool
+        except:
```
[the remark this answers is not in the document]
imagoulas: It is exactly what you said. At the moment, symmetry is supported only for `system_type = molecule`. If someone uses `system_type = model` there will be no `orb_irreps_to_int` attribute in `self._sys`. Can you please clarify what you mean by "specifying exactly which kind of exception"? Should I just put a comment in the code explaining why this is exception is needed or have a print statement alerting the user that symmetry is not supported for `system_type = model`?

src/qforte/adapters/molecule_adapters.py
```diff
@@ -131,6 +136,9 @@ def create_openfermion_mol(**kwargs):
 
     openfermion_mol.load()
 
+    # Get symmetry information
+    point_group, orb_irreps, orb_irreps_to_int = psi4_symmetry(kwargs['filename']+'.out')
```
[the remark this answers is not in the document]
imagoulas: This is for two reasons.
1) For the sake of "symmetry" between how we treat `build_type = 'psi4'` and `build_type = 'openfermion'`. The C1 symmetry is still a point group after all :-).
2) If we decide to keep the interface to OpenFermion, they might allow point group symmetry in the future. Our code would be ready for such an update.

src/qforte/abc/algorithm.py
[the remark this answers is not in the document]
JonathonMisiewicz: No, I'm referring to [a feature of the Python language](https://docs.python.org/3/tutorial/errors.html#handling-exceptions). line 272 should be `except AttributeError:` so that only that kind of error is caught. You should also have a comment in the code to specify why an `AttributeError` means no symmetry.

src/qforte/utils/point_groups.py
imagoulas: The initial name of the function was something along the lines: `excit_symmetry_check`. The reason that I made it more vague is that in UCC we have both excitation and de-excitation operators. I can call it `sq_op_symmetry_check`. What do you think?

src/qforte/utils/point_groups.py
imagoulas: I know that by default the function will return `None`, I just thought that putting it there explicitly made the code more transparent. I will remove it.

src/qforte/utils/point_groups.py
JonathonMisiewicz: With the new information you've given me, my objection is to the word "check". We're not _checking_; we're _finding_. Adding `sq_op` is good.

JonathonMisiewicz (changes requested):
Scraping data from an output file is usually a bad idea for long-term code. It's fine for personal scripts, but we need to adjust our scraper any time Psi changes their output file.

A better course of action is to read the attributes we need directly from [Psi's wavefunction object](https://psicode.org/psi4manual/master/api/psi4.core.Wavefunction.html?highlight=wavefunction#psi4.core.Wavefunction). In the likely event that you can't figure out how to get information from this wavefunction, let me know, and I'll update the docs, Psi-side.

Of course, this doesn't help with OpenFermion, but if OpenFermion supports symmetry, we'll need to write an interface to them, too. The better course of action is to have a special function to define attributes for `c1` symmetry.

src/qforte/utils/point_groups.py
JonathonMisiewicz: In this case, _not_ specifying a return value makes explicit that the point of the code isn't about a return value, but about something else. In this case, printing.

imagoulas:
I couldn't agree with you more. Please believe me when I say that I tried really hard to figure out how to extract the irreducible representations of each orbital from Psi's wavefunction object, but, alas, I failed. That's why I resorted to extracting the information from the output. If you can give me some advice in this direction I would greatly appreciate it. In this scenario, I will extract the information directly from the wavefunction object and remove the "symmetry" from OpenFermion, since we do not have direct access to that information, I believe.

JonathonMisiewicz:
Okay, I can make an issue in the Psi docs if you tell me what piece you couldn't find.

The wavefunction has `wfn.Ca()` and `wfn.Cb()`. You can access irrep `n` by the NumPy syntax described [here](https://psicode.org/psi4manual/master/numpy.html). The columns of irrep `n` index MOs of that irrep.

imagoulas:
The Psi4 manual was clear enough that I was already aware of the `wfn.Ca()` business and the Numpy syntax. My trouble was that I could not figure out how this information would be useful for what we wanted to accomplish. `wfn.Ca().nph` prints the definition of MOs in terms of AOs, with the MOs of a given irrep grouped together. Although this is useful for finding how many MOs belong to each irrep, I could not order the irreps in ascending orbital energies. I just figured out how to do that; I will use the `wfn.epsilon_a().nph` command. This prints the energies of the MOs, with the MOs of a given irrep grouped together. I will see what I can still do today, but most likely I will submit the changes tomorrow.

JonathonMisiewicz:
Please add pytests for new symmetry features, so I know what's expected to work now.

imagoulas:
I added four different tests. The first one only checks that the various symmetry attributes are set correctly in the Molecule object. The other three tests check that the symmetry-adapted SQOpPools have the expected sizes for UCCNVQE, ADAPTVQE, and UCCNPQE. I could combine these tests into a single one by looping over the three methods. Although this could potentially speed up the time it takes for the test to run, I decided not to do it for clarity. If you think that looping over the three methods is acceptable, let me know and I will push a new commit.

src/qforte/ucc/uccnvqe.py
```diff
@@ -75,8 +77,8 @@ def run(self,
 
         self.fill_pool()
 
-        if self._verbose:
-            self._pool_obj.print_pool()
+        #if self._verbose:
+        #    self._pool_obj.print_pool()
```
JonathonMisiewicz, line 81:
```suggestion
        if self._verbose:
            print(self._pool_obj.str())
```

src/qforte/ucc/uccnvqe.py
```diff
@@ -53,6 +54,7 @@ def run(self,
         self._optimizer = optimizer
         self._pool_type = pool_type
         self._noise_factor = noise_factor
+        self._irrep = irrep
```
JonathonMisiewicz, line 57: `_irrep` should not be a property of the `Ansatz`. It should be a property of `System`, which any `Ansatz` has. Please remove on all three subclasses and change the population of the operator pool accordingly.

tests/symmetry_tests.py
```diff
@@ -0,0 +1,206 @@
...
+
+            self.assertEqual(mol.point_group, [group, irreps[count]])
+            self.assertEqual(mol.orb_irreps, orb_irreps[count])
+            self.assertEqual(mol.orb_irreps_to_int, orb_irreps_to_int[count])
+
+            with patch('sys.stdout', new = StringIO()) as fake_out:
+                char_table([group, irreps[count]])
+                self.assertEqual(fake_out.getvalue(), char_tables[count])
+
+    def test_symmetry_uccnvqe(self):
```
JonathonMisiewicz, line 124: It's clearer to combine this test and the two below into a single test. Please use the `param` feature of `pytest`. Example [here](https://github.com/psi4/psi4/blob/master/tests/pytests/test_dipoles.py).

src/qforte/adapters/molecule_adapters.py
```diff
@@ -279,6 +282,25 @@ def create_psi_mol(**kwargs):
     # run psi4 caclulation
     p4_Escf, p4_wfn = psi4.energy('SCF', return_wfn=True)
 
+    # Get symmetry information
+    orbitals = []
+    for irrep in range(len(p4_wfn.epsilon_a().nph)):
+        for orbital in p4_wfn.epsilon_a().nph[irrep]:
```
JonathonMisiewicz, line 288:
```suggestion
    for irrep in p4_wfn.epsilon_a().nph:
        for orbital in irrep:
```

src/qforte/abc/algorithm.py
```diff
@@ -258,6 +260,21 @@ def fill_pool(self):
         else:
             raise ValueError('Invalid operator pool type specified.')
 
+        # If possible, impose symmetry restriction to operator pool
```
JonathonMisiewicz, line 263: I have doubts about whether this code belongs in the `Algorithm` class, but I can deal with this later. A PR I have planned will clarify the matter.

JonathonMisiewicz (changes requested):

src/qforte/utils/point_groups.py
```diff
@@ -0,0 +1,69 @@
...
+    point_group: list of two lists; point_group[0] holds the name of the group
+                                    point_group[1] is a list that holds the irreps of the group
+
+    """
+
+    if type(point_group) != list:
+        raise TypeError("""{0} is not a list.
+                This function takes arguments of the form:\n
+                [['c2v'], ['A1', 'A2', 'B1', 'B2']]]\n
+                using the so-called Cotton ordering of the irreps.""".format(type(point_group)))
```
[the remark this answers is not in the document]
imagoulas: The extra `\n` are deliberate. I want to have empty lines above and below the example.

src/qforte/adapters/molecule_adapters.py
imagoulas: This is not correct. The way I am doing it creates a list with elements of the form `[orbital energy, orbital irrep index]`. This allows me to sort the orbital irreps according to the orbital energies. What you are suggesting creates a list with elements of the form `[orbital energy, numpy array of orbital energies of a given irrep]`, which is not what we need. I will stick with the original version of the code.

src/qforte/adapters/molecule_adapters.py
```diff
@@ -279,6 +282,25 @@ def create_psi_mol(**kwargs):
...
+            orbitals.append([orbital, irrep])
```
JonathonMisiewicz, line 289:
```suggestion
    for irrep, block in enumerate(p4_wfn.epsilon_a().nph):
        for orbital in block:
            orbitals.append([orbital, irrep])
```

src/qforte/ucc/uccnvqe.py
imagoulas: I do not necessarily agree with this. The irreducible representation of a particular state is not a property of the system, in our case molecule. A given molecule, at a given geometry, belongs to a given point group, but nothing can be said about an irreducible representation, unless we commit to a given state. The way I see it is as follows. The system class defines our Fermi vacuum, which for closed-shell species translates to a totally symmetric state (I believe that, for the time being, QForte does not work for open-shell molecules, but please correct me if I am wrong). If we want to study states that belong to a different irreducible representation than the RHF Slater determinant, we need to construct an Ansatz that belongs to the irreducible representation that we want. For example, in GAMESS, and I am pretty sure that this is how it is done in Psi4 as well, if one wants to perform an FCI calculation they need to provide the irreducible representation of the target state (if nothing is specified, the default is totally symmetric). This is the reason why `_irrep` should be an attribute of the Ansatz and not the `System`.

src/qforte/adapters/molecule_adapters.py
imagoulas: This one works!

src/qforte/utils/point_groups.py
```diff
@@ -0,0 +1,69 @@
...
+objects with "build_type = 'psi4'".
+"""
+
+def char_table(point_group):
+    """
+    Function that prints the character table of a chosen point group.
+
+    Parameters
+    ----------
+    point_group: list of two lists; point_group[0] holds the name of the group
```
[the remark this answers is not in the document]
imagoulas: You are correct. I phrased it very badly. `point_group` is a list of two elements. The first element is a string containing the point group symbol and the second element is a list containing the irreps of the point group, according to the Cotton ordering.

src/qforte/ucc/uccnvqe.py
JonathonMisiewicz: Right, I was reviewing this a little too quickly. `_irrep_` should definitely not be a property of `System`.

I still do not want `_irrep` as a member of the `Ansatz` class. `_irrep` makes perfect sense for ansatze expressed in terms of fermionic second quantization, but not for an ansatz expressed in terms of Pauli strings. What about having a member called `symmetries` or `states` that is a dict to store all the relevant symmetries? For molecular systems, we can have a key called `_irrep` that hashes to the irrep of the state of interest.

imagoulas: I like this idea. Eventually, I could add the information about the Z2 symmetries there as well. I just want to confirm that I understand this correctly, so that I can implement it. Symmetries/states would be a class of its own or an attribute to an existing class? The word "member" confuses me a little :-)

JonathonMisiewicz: Attribute of an existing class.

src/qforte/abc/algorithm.py
imagoulas: Ideally, this code should have been part of the C++ function `fill_pool`. The way the code works now is not the most efficient one: first the `SQOpPool` is filled in the C++ side of the code, without any symmetry restrictions, and then symmetry is imposed in Python. Placing this code in the C++ function `fill_pool` will also allow us to remove the `try` and `except` statements, since the line `elif isinstance(self._pool_type, qf.SQOpPool):` does not involve the C++ `fill_pool` function. To be able to do this, we would need to pass the information regarding orbital irreps from Python to C++. My current knowledge (or lack thereof) of both Python and C++ prevents me from doing this. If you feel that this is the best solution, I would be happy to discuss how to proceed.

JonathonMisiewicz: Let's save that for a future PR.

src/qforte/abc/algorithm.py
```diff
@@ -249,6 +250,8 @@ def fill_pool(self):
         """ This function populates an operator pool with SQOperator objects.
         """
 
+        from qforte import sq_op_find_symmetry
```
JonathonMisiewicz, line 253: All `import` statements should be moved to the top of the file.

src/qforte/abc/algorithm.py
```diff
@@ -293,6 +307,26 @@ def __init__(self, *args, **kwargs):
         self._pool = []
         self._pool_obj = qf.SQOpPool()
 
+        kwargs.setdefault('irrep', None)
+        if hasattr(self._sys, 'point_group'):
+            irreps = [x for x in range(len(self._sys.point_group[1]))]
+            if kwargs['irrep'] == None:
```
JonathonMisiewicz, line 313:
```suggestion
            if kwargs['irrep'] is None:
```

src/qforte/abc/algorithm.py
```diff
@@ -293,6 +307,26 @@ def __init__(self, *args, **kwargs):
...
+            elif kwargs['irrep'] in irreps:
+                self._irrep = kwargs['irrep']
+            else:
+                raise ValueError("{0} is not an irreducible representation of {1}.\n"
+                                 "               Choose one of {2} corresponding to the\n"
+                                 "               {3} irreducible representations of {1}".format(kwargs['irrep'],
+                                     self._sys.point_group[0].capitalize(),
+                                     irreps,
+                                     self._sys.point_group[1]))
+        elif kwargs['irrep'] != None:
```
JonathonMisiewicz, line 326:
```suggestion
        elif kwargs['irrep'] is not None:
```

src/qforte/abc/algorithm.py
```diff
@@ -293,6 +307,26 @@ def __init__(self, *args, **kwargs):
         self._pool = []
         self._pool_obj = qf.SQOpPool()
 
+        kwargs.setdefault('irrep', None)
+        if hasattr(self._sys, 'point_group'):
+            irreps = [x for x in range(len(self._sys.point_group[1]))]
```
JonathonMisiewicz, line 312:
```suggestion
            irreps = list(range(len(self._sys.point_group[1])))
```

src/qforte/system/molecular_info.py
```diff
@@ -162,3 +162,34 @@ def ccsd_energy(self):
...
+    @property
+    def orb_irreps_to_int(self):
+        return self._orb_irreps_to_int
+
+    @orb_irreps_to_int.setter
+    def orb_irreps_to_int(self, orb_irreps_to_int):
+        self._orb_irreps_to_int = orb_irreps_to_int
+
+    @property
+    def hf_orbital_energies(self):
```
JonathonMisiewicz, line 190: Why exactly does `hf_orbital_energies` exist? Was this added for a feature that was scrapped from this PR?

src/qforte/system/molecular_info.py
```diff
@@ -162,3 +162,34 @@ def ccsd_energy(self):
     def ccsd_energy(self, ccsd_energy):
         self._ccsd_energy = ccsd_energy
 
+    @property
```
JonathonMisiewicz, line 165: Please define what these variables are supposed to be (especially expected types) in a docstring.

src/qforte/system/molecular_info.py
imagoulas: The hf_orbital_energies attribute is just a "placeholder" to facilitate a future PR. The intention is to extract the orbital energies directly from Psi4 instead of using QForte to calculate them as expectation values. Before we are able to do so, we need to deal with two issues.
1) We need to see how to extract orbital energies from OpenFermion (or get rid of OpenFermion altogether)
2) We need to write an HF code for model Hamiltonians to extract the orbital energies (or keep OpenFermion if such an option already exists).

src/qforte/abc/algorithm.py
imagoulas: If I move the import statements to the top of the file wouldn't sq_op_find_symmetry be imported even in situations where we do not needed? I thought that placing it inside the particular function that requires it would help with memory and, perhaps, speeding up the code.

src/qforte/system/molecular_info.py
JonathonMisiewicz: While _I'd_ prefer splitting `hf_orbital_energies` into a separate PR, because it's unrelated to "adding point group symmetry," fine.

I'm personally in favor of dropping OpenFermion support.

src/qforte/abc/algorithm.py
JonathonMisiewicz: Yes, it may  imported even when not needed, but that's not a problem. Performance impact of `sq_op_find_symmetry` is likely negligible.

src/qforte/system/molecular_info.py
imagoulas: I am also in favor of dropping OpenFermion. After this PR, I can work on removing the dependency.

JonathonMisiewicz (approved):
Merge when you feel it's ready.

imagoulas:
I realized that if I run `pytest` in the `tests` directory, the `symmetry_tests.py` file was not picked up by pytest. A quick survey of the literature (googling for a few minutes) revealed that pytest assumes that all files are of the form `test_*.py`. Renaming `symmetry_tests.py` to `test_symmetry.py` fixed the issue. I also placed the two tests contained in test_symmetry.py inside a class, to be consistent with the other test files. Since this is a trivial modification, I will go ahead and merge the PR. I also discovered, at least on my computer, that running `pytest *.py` checks 45 tests while simply running `pytest` checks 42. My next PR will replace unittest by pytest for all tests, updating also the readme instructions. This also means that I will rename all tests as described above.

PR #139, opened 2021-09-16
Replaced unittest by pytest in tests.

imagoulas:
## Description
I replaced the use of `unittest` by `pytest` in the test files. I renamed all test files so that they look like `test_*.py`. I also renamed all classes in the test files so that they look like `Test*`. This allows pytest to pick up all test files when one runs `pytest` in the `tests` folder. This information is also reflected in the updated `README` file. To remove the annoying warning messages regarding unrecognized options of the minimize function of scipy, I made a few minor changes to `uccnvqe.py`.

All tests passed successfully without any warnings. Nevertheless, it would be great if @JonathonMisiewicz you could also confirm this at your end.

## Checklist
- [x ] Added/updated tests
- [x ] Ready to go!


tests/test_circ_ref.py
```diff
@@ -39,10 +32,4 @@ def test_srqk_with_uccsdpqe_ref(self):
...
 
         Egs = alg2.get_gs_energy()
-        self.assertLess(abs(Egs-Efci), 5.0e-6)
-
-
-
-
-if __name__ == '__main__':
-    unittest.main()
+        assert Egs == approx(Efci, abs=5.0e-7)
```
JonathonMisiewicz, line 35:
```suggestion
        assert Egs == approx(Efci, abs=5.0e-6)
```

Merge after you change this back.

JonathonMisiewicz (approved):

PR #141, opened 2021-09-16
Removed OpenFermion Support

imagoulas:
## Description
Removed support for OpenFermion, since we have a direct interface to Psi4. In doing so, I updated the `README` file with new installation instructions and also `.travis.yml`. Two attributes (`hdf5_dir`, `description`) that were not used when build_type='psi4' were removed.

By using the latest version of python, a deprecation warning was triggered during QForte compilation. The warning was concerned with the `PyEval_InitThreads()` function in pybind11. After a careful examination of the literature (googling for a few minutes), I found a way to solve this by making a slight modification in the `internals.h` file of pybind11 (see https://github.com/knarfS/smuview/commit/9f0812925fa90997e71c703c9330f1c172211359 and search for PyEval_InitThreads()).

All tests passed successfully and without any warnings.

## Checklist
- [x ] Checked for redundant headers/imports
- [x ] Ready to go!


lib/pybind11/include/pybind11/detail/internals.h
```diff
@@ -191,7 +191,9 @@ PYBIND11_NOINLINE inline internals &get_internals() {
         auto *&internals_ptr = *internals_pp;
         internals_ptr = new internals();
 #if defined(WITH_THREAD)
-        PyEval_InitThreads();
+        #if PY_VERSION_HEX < 0x03090000
```
JonathonMisiewicz, line 194: What is this code doing?

README.md
```diff
@@ -12,13 +12,14 @@ Install Dependencies (Recommended)
 
 #### create and activate qforte environment:
 ```bash
-conda create -n qforte_env python=3.7
```
JonathonMisiewicz, line 15: I understand removing `openfermion` and `openfermionpsi4`. What motivates the other changes?

lib/pybind11/include/pybind11/detail/internals.h
imagoulas: Short story: I do not know.
Long story: According to docs.python.org "In Python 3.6 and older, this function created the GIL if it didn’t exist". In Python 3.9 the `PyEval_InitThreads()` command is deprecated and it will be completely removed by Python 3.11. This code is checking the version of Python. If it is less than 3.9, `PyEval_InitThreads()` is executed, otherwise it is not. By putting this if statement I could compile QForte using Python 3.9 without any deprecation warning messages.

JonathonMisiewicz: Do _not_ edit the code in `lib`.

Those are external libraries used by QForte - we shouldn't change them. What QForte currently does (including its own copy of those libraries) is bad practice. Making this disagree from the standard versions is worse practice.

README.md
imagoulas: The reason we were requesting Python 3.7 was the OpenFermion dependency. Since we do not support OpenFermion anymore there is no reason to request for an older version of python. The problem of Psi4 not installing the latest version was also caused by Python 3.7. Using the latest version of Python, there is, as of now, no need to specify a particular version of Psi4.

When not installing OpenFermion and OpenFermion-Psi4, I got a bunch of errors when running the tests claiming that scipy was not recognized. I guess that scipy was installed as a dependency of OpenFermion. Now that we do not install OpenFermion, we need to install scipy. All of these changes stem from a series of trial and error experiments.

lib/pybind11/include/pybind11/detail/internals.h
imagoulas: I do understand that this is an external library. The change that I made is pretty innocent, but I will revert back. As you imply, we shouldn't be providing our own copy, because codes get updated. In a future PR, we should fix this.

JonathonMisiewicz (approved):

Issue #144, opened 2021-09-17
Synchronize `_pool_obj` and `_pool`


JonathonMisiewicz:
The `AnsatzAlgorithm` class has both a [`_pool_obj` and a `_pool`](https://github.com/evangelistalab/qforte/blob/master/src/qforte/abc/algorithm.py#L306-L307), where `_pool` _should_ just be the `_pool_obj`'s list. This is redundant, and `_pool` should be eliminated.

- [ ] Synchronize `_pool_obj.terms_` and `_pool`... Currently de-sync'ed in PQE
- [ ] Add magic methods to `SQOpPool`
    - [x] `__getitem__`
    - [x] `__len__`
    - [x] `__iter__`
- [ ] Remove all current calls to `_pool` with calls to `_pool_obj` (using the magic methods) or `_pool_obj.terms()`
- [ ] Eliminate the `_pool` member of AnsatzAlgorithm

PR #145, opened 2021-09-17
Define SQOpPool.__len__

JonathonMisiewicz:
## Description
Binds a `__len__` magic method so that `len(SQOpPool)` is the number of terms in the pool. This finishes item 2B of #144 and accomplishes part of item 3.

## User Notes
- [x] Defined `len(SQOpPool)`

## Checklist
- [x] Ready to go!


imagoulas (approved):
Approved and merged.

PR #150, opened 2021-09-21
Use ansatz_circuit

JonathonMisiewicz:
## Description
Eliminates some code duplication by invoking `ansatz_circuit`. There's more repetition to remove, but this reduces the number of `_pool`/`_pool_obj` calls I need to track, which is what I need right now.

Also removes the unused `_commutator_pool` attribute of `SPQE`.

## Checklist
- [x] Tests pass
- [x] Ready to go!


imagoulas (approved):
Everything looks good to me. Good catch about the code repetition. Approved and merged!

PR #155, opened 2021-09-22
Added feature to find the Z2 symmetries of the Hamiltonian.

imagoulas:
## Description
This is the first step toward implementing qubit tapering in QForte. The new functions can find the Z2 symmetries of the Hamiltonian of interest, compute the Clifford gates, and taper off qubits from a given qubit operator and reference state.

The new test_qubit_tapering.py file checks that the symmetries found by QForte and Qiskit for H2O/STO-3G are identical. Once additional functionalities pertaining to qubit tapering become available, more tests will be added to this file.

The next step will be to interface these functions with the existing algorithms in QForte, starting, for example, with UCCNVQE.

## User Notes
- [x] Features added

## Checklist
- [x] Documented source code
- [x] All tests passed successfully
- [x] Ready to go!

### DIFF
--- a/src/qforte/utils/qbt_tapering.py
+++ b/src/qforte/utils/qbt_tapering.py
@@ -265,16 +265,23 @@ def qubit_tapering_oprtr(tapered_qubits, sign, operator, unitary):
 
     # Similarity transform the given operator using the fact that the unitary matrix is also Hermitian
 
+    ## This is necessary; otherwise later on we would be modifying the "unitary" object directly, which we do not want to do
     unitary_tmp = qf.QubitOperator()
     unitary_tmp.add_op(unitary)
-    operator.operator_product(unitary_tmp, True, True)
-    unitary_tmp.operator_product(operator, True, True)
-    operator = unitary_tmp
+
+    ## This is necessary; if we do not do this, when transforming the Hamiltonian we would be changing the "hamiltonian"
+    ## attribute of the Molecule class to nonsense.
+    operator_tmp = qf.QubitOperator()
+    operator_tmp.add_op(operator)
+
+    operator_tmp.operator_product(unitary_tmp, True, True)
+    unitary_tmp.operator_product(operator_tmp, True, True)
+    operator_tmp = unitary_tmp
 
     # Create a QForte QubitOperator object that will store the tapered operator
     operator_tapered = qf.QubitOperator()
 
-    for i, (coeff, circuit) in enumerate(operator.terms()):
+    for i, (coeff, circuit) in enumerate(operator_tmp.terms()):
         # Create circuit that will store the pauli string with gates corresponding to qubits to be tapered off removed
         circ = qf.Circuit()
         # List that will store the coefficients modified according to the sign structure provided by the user
--- a/src/qforte/utils/qbt_tapering.py
+++ b/src/qforte/utils/qbt_tapering.py
@@ -210,7 +210,7 @@ def find_Z2_symmetries(hamiltonian, qiskit_order = False):
 
     reorder = np.argsort(sigma_x)
     sigma_x = sigma_x[reorder][::-1]
-    gnrtrs = [gnrtrs[i] for i in reorder]
+    gnrtrs = [gnrtrs[i] for i in reorder[::-1]]
 
     ## Construct individual unitary matrices
     untrs = []
--- a/src/qforte/utils/qbt_tapering.py
+++ b/src/qforte/utils/qbt_tapering.py
@@ -1,0 +1,318 @@
+"""
+Functions for performing qubit tapering based on the Z2 symmetries of the qubit Hamiltonian
+"""
+
+import qforte as qf
+import numpy as np
+from itertools import product
+
+def find_Z2_symmetries(hamiltonian, qiskit_order = False):
+
+    """
+    This function computes the Z2 symmetries of the Hamiltonian, identifies the ID numbers of the qubits to
+    be tapered off, provides the corresponding sign structures, and constructs the necessary unitary operators.
+    It is based on the qubit tapering approach reported in https://arxiv.org/abs/1701.08213.
+
+    Arguments:
+
+    hamiltonian: QubitOperator
+        The qubit Hamiltonian whose Z2 symmetries we want to find.
+
+    qiskit_order: bool
+        If True, the qubits to be tapered off will have the smallest id numbers possible, as is done in Qiskit.
+        If False, the qubits to be tapered off will have the largest id numbers possible. This is the default,
+        since it leads to less operations once the qubits are removed.
+    """
+
+    # The algorithm consists of the following steps:
+    # 1) Convert qubit Hamiltonian to binary matrix, called the parity check matrix E
+    # 2) Find the basis of the kernel of the parity check matrix, ker{E}
+    # 3) Find maximal Abelian subgroup of the basis of ker{E}
+    # 4) Construct unitary matrices that will be used in the qubit_tapering algorithm
+    # 5) Compute the various signs corresponding to the eigenvalues of the sigma_x operators that will be removed
+
+    # 1) Convert qubit Hamiltonian to binary matrix
+
+    ## Obtain number of qubits and number of Pauli strings contained in qubit Hamiltonian
+    n_qubits = hamiltonian.num_qubits()
+    n_strings = len(hamiltonian.terms())
+
+    ## Construct parity check matrix
+    prt_chck_mtrx = np.zeros((n_strings, 2 * n_qubits), dtype=np.uint8)
+    for i, string in enumerate(hamiltonian.terms()):
+        for pauli in string[1].gates():
+            XYZ_gate = pauli.gate_id()
+            trgt = pauli.target()
+            if XYZ_gate == 'Z':
+                prt_chck_mtrx[i, n_qubits - trgt - 1] = 1
+            elif XYZ_gate == 'X':
+                prt_chck_mtrx[i, 2 * n_qubits - trgt - 1] = 1
+            else:
+                prt_chck_mtrx[i, n_qubits - trgt - 1] = 1
+                prt_chck_mtrx[i, 2 * n_qubits - trgt - 1] = 1
+
+    # 2) Find the basis of the kernel of the parity check matrix, ker{E}
+    # See https://en.wikipedia.org/wiki/Kernel_(linear_algebra)#Computation_by_Gaussian_elimination
+
+    ## Augment the (n_strings) × (2 * n_qubits) parity check matrix by the (2 * n_qubits) × (2 * n_qubits) identity matrix
+
+    prt_chck_mtrx_aug = np.concatenate((prt_chck_mtrx, np.identity(2 * n_qubits, dtype=np.uint8)), axis=0)
+
+    ## Find column echelon form (CEF) of augmented matix.
+
+    idx = 0
+    for row in range(n_strings):
+        ### Find a non-zero row
+        if np.any(prt_chck_mtrx_aug[row,idx:] == 1):
+            ### Find positions of non-zero elements
+            count = np.argwhere(prt_chck_mtrx_aug[row,idx:] == 1)
+            column = int(count[0]) + idx
+            ### Swap columns to bring first non-zero element in this row immediately to the right of that of the previous row
+            prt_chck_mtrx_aug[:, [column , idx]] = prt_chck_mtrx_aug[:, [idx, column]]
+            ### Eliminate remaining non-zero elements in the submatrix
+            for i in range(1,count.shape[0]):
+                column2 = int(count[i]) + idx
+                prt_chck_mtrx_aug[:,column2] = np.bitwise_xor(prt_chck_mtrx_aug[:, idx], prt_chck_mtrx_aug[:,column2])
+            ### decrease the number of columns that need to be checked
+            idx += 1
+
+    ## Find basis vectors of kernel by checking for zero columns in the CEF of the parity check matrix
+
+    dim = 0
+    basis = np.zeros((1, 2 * n_qubits), dtype=np.uint8)
+    for column in range(2 * n_qubits):
+        row = n_strings
+        if not np.any(prt_chck_mtrx_aug[:row, column]):
+            if dim == 0:
+                basis = prt_chck_mtrx_aug[row:, column].reshape((1,2 * n_qubits))
+                dim += 1
+            else:
+                basis = np.concatenate((basis, prt_chck_mtrx_aug[row:, column].reshape((1,2 * n_qubits))), axis=0)
+
+    # 3) Find maximal Abelian subgroup of the basis of ker{E}
+    # This is done using the symplectic bilinear form B that the binary vector space is equipped with:
+    # B(v,w) = v * B * w.T,
+    # where v and w are binary vectors and B is the matrix of the bilinear form.
+
+    ## Construct matrix of bilinear form.
+
+    top = np.concatenate((np.zeros((n_qubits, n_qubits), dtype=np.uint8), np.identity(n_qubits, dtype=np.uint8)), axis=1)
+    bottom = np.concatenate((np.identity(n_qubits, dtype=np.uint8), np.zeros((n_qubits, n_qubits), dtype=np.uint8)), axis=1)
+    blnr_frm_mtrx = np.concatenate((top, bottom), axis=0)
+
+    ## Construct abln matrix.
+    ## abln[i,j] contains the value of the bilinear form betweeen the ith and jth basis vectors of ker{E}
+    ## A "0"/"1" in the ith row and jth column of abln means that the Pauli strings
+    ## associated with the ith and jth binary vectors commute/anticommute.
+
+    aux = np.matmul(blnr_frm_mtrx, basis.T)
+
+    abln = np.zeros((basis.shape[0], basis.shape[0]), dtype=np.uint8)
+    for row in range(basis.shape[0]):
+        for column in range(basis.shape[0]):
+            for idx in range(2*n_qubits):
+                abln[row, column] = (abln[row, column] + basis[row, idx] * aux[idx, column]) % 2
+
+    ## Find the maximal abelian subgroup iteratively. This is done as follows:
+    ## 1) Find non-zero row of abln with smallest Hamming weight.
+    ##    The corresponding basis vector e commutes with most other basis vectors.
+    ## 2) Find basis vectors that do not commute with e and remove them.
+    ## 3) Reconstruct abln matrix with remaining basis vectors.
+    ## 4) If abln == 0 exit otherwise go to step 1) and repeat.
+
+    basis_tmp = basis
+
+    while not np.all(abln == 0):
+
+        ### Maximum possible Hamming weight of a given binary vector
+        hmng = 2 * n_qubits
+
+        idx = 0
+        for row in range(abln.shape[0]):
+            ### Find the basis vector e that commutes with most of the remaining basis vectors
+            hmng_tmp = np.sum(abln[row, :])
+            if hmng_tmp > 0 and hmng > hmng_tmp:
+                hmng = hmng_tmp
+                idx = row
+
+        ### Identify basis vectors that do not commute with e and eliminate them.
+        count = np.argwhere(abln[idx,:] == 1)
+        basis_tmp = np.delete(basis_tmp, count, axis=0)
+
+        ### Construct abln matrix for the next iteration.
+
+        aux = np.matmul(blnr_frm_mtrx, basis_tmp.T)
+
+        abln = np.zeros((basis_tmp.shape[0], basis_tmp.shape[0]), dtype=np.uint8)
+        for row in range(basis_tmp.shape[0]):
+            for column in range(basis_tmp.shape[0]):
+                for idx in range(2*n_qubits):
+                    abln[row, column] = (abln[row, column] + basis_tmp[row, idx] * aux[idx, column]) % 2
+
+    tau = basis_tmp
+
+    #### This step is not necessary. It transforms the generators of the symmetry group in such a way that the
+    #### removed qubits have the smallest/largest possbile identity numbers. A similar modification is made
+    #### at the "sigma_x selection" part of the algorithm below. The "smallest" case produces identical resutls
+    #### with those obtained using Qiskit, after taking into account that in Qiskit Slater determinants are
+    #### represented by strings of alpha orbitals followed by the beta ones.
+    idx = 0
+    lst = []
+    if qiskit_order:
+        _range = list(reversed(range(n_qubits, 2 * n_qubits)))
+    else:
+        _range = list(range(n_qubits, 2 * n_qubits))
+    for column in _range:
+        count = np.argwhere(tau[:, column] == 1)
+        if count.shape[0] == 1:
+            if count[0] not in lst:
+                lst.append(int(count[0]))
+                idx += 1
+        else:
+            for i in range(count.shape[0]):
+                if count[i] not in lst:
+                    lst.append(int(count[i]))
+                    for el in filter(lambda el: el not in count[i], count):
+                        tau[int(el), :] = np.bitwise_xor(tau[int(el), :], tau[int(count[i]), :])
+                    idx += 1
+                    break
+        if idx == tau.shape[0]:
+            break
+    ####
+
+    # 4) Construct unitary matrix that will transform Hamiltonian
+
+    ## Translate binary vectors back to Pauli strings to obtain the generators of the symmetry group
+    gnrtrs = []
+    for i in range(tau.shape[0]):
+        pauli = qf.Circuit()
+        for j in reversed(range(n_qubits)):
+            if tau[i, j] == 1 and tau[i, j + n_qubits] == 1:
+                pauli.add_gate(qf.gate('Y', n_qubits - j - 1))
+            elif tau[i, j] == 1:
+                pauli.add_gate(qf.gate('X', n_qubits - j - 1))
+            elif tau[i, j + n_qubits] == 1:
+                pauli.add_gate(qf.gate('Z', n_qubits - j - 1))
+        gnrtrs.append(pauli)
+
+    ## Find which sigma_x gate will be paired with which generator
+    sigma_x = np.zeros((tau.shape[0]), dtype=np.uint16)
+
+    if qiskit_order:
+        _range = list(range(n_qubits))
+    else:
+        _range = list(reversed(range(n_qubits)))
+    for i in _range:
+        if np.argwhere(tau[:,n_qubits + i] == 1).shape[0] == 1:
+            sigma_x[np.argwhere(tau[:,n_qubits + i] == 1)] = n_qubits - i - 1
+
+    ## Reorder sigma_x operators in descending order of qubit identity. Do the same with the generators for consistency.
+
+    reorder = np.argsort(sigma_x)
+    sigma_x = sigma_x[reorder][::-1]
+    gnrtrs = [gnrtrs[i] for i in reorder]
+
+    ## Construct individual unitary matrices
+    untrs = []
+
+    for idx, generator in enumerate(gnrtrs):
+        circ = qf.Circuit()
+        circ.add_gate(qf.gate('X', sigma_x[idx]))
+        oprtr = qf.QubitOperator()
+        oprtr.add_term(1/np.sqrt(2), generator)
+        oprtr.add_term(1/np.sqrt(2), circ)
+        untrs.append(oprtr)
+
+    ## Construct final unitary matrix
+    unitary = qf.QubitOperator()
+    unitary.add_op(untrs[0])
+    for op in untrs[1:]:
+        unitary.operator_product(op, True, True)
+
+    # 5) Compute the various signs corresponding to the eigenvalues of the sigma_x operators that will be removed
+
+    signs = list(product([1, -1], repeat=len(sigma_x)))
+
+    return gnrtrs, sigma_x, untrs, unitary, signs
+
+def qubit_tapering_oprtr(n_qubits, tapered_qubits, sign, operator, unitary):
+
+    """
+    This function uses the unitary matrix obtained using find_Z2_symmetries to transform a given operator.
+    After the unitary transformation, the resulting operator acts on the qubits to be tapered off by at most
+    a single sigma_x operator. The sigma_x gates acting on qubits to be tapered off are replaced by their
+    +/- 1 eigenvalues, which are provided by the user.
+
+    Arguments
+
+    n_qubits: integer
+        Initial number of qubits.
+
+    tapered_qubits: list of integers
+        The ID numbers of the qubits to be tapered off.
+
+    sign: list of +/- 1.
+        The desired sign structure corresponding to the eigenvalues of the sigma_x operators acting on
+        the qubits to be tapered off.
+
+    operator: QubitOperator
+        The operator that we wish to transform. For example, the qubit Hamiltonian or the UCC ansatz.
+
+    unitary: QubitOperator
+        The unitary operator that will perform the dersired transformation.
+    """
+
+    # Transform the given operator using the fact that the unitary matrix is also Hermitian
+
+    unitary_tmp = qf.QubitOperator()
+    unitary_tmp.add_op(unitary)
+    operator.operator_product(unitary_tmp, True, True)
+    unitary_tmp.operator_product(operator, True, True)
+    operator = unitary_tmp
+
+    # Create a QForte QubitOperator object that will store the tapered operator
+    operator_tapered = qf.QubitOperator()
+
+    coeff_tmp = []
+    for counter, term in enumerate(operator.terms()):
+        # Create circuit that will store the pauli string with gates corresponding to qubits to be tapered off removed
+        circ = qf.Circuit()
+        # List that will store the coefficients modified according to the sign structure provided by the user
+        coeff_tmp.append(term[0])
+        for pauli in term[1].gates():
+            if any(qubit == pauli.target() for qubit in tapered_qubits):
+                coeff_tmp[counter] *= sign[int(np.argwhere(tapered_qubits == pauli.target()))]
+            else:
+                # Adjust the qubit ID of a given gate based on the number of gates that were removed before it
+                idx = np.count_nonzero(pauli.target() > tapered_qubits)
+                target = pauli.target() - idx
+                gate = pauli.gate_id()
+                circ.add_gate(qf.gate(gate, target))
+
+        operator_tapered.add_term(term[0], circ)
+
+    # Replace the coefficients with the ones having the requested sign structure
+    operator_tapered.set_coeffs(coeff_tmp)
+    # Simplify the entire operator
+    operator_tapered.simplify(True)
+
+    return operator_tapered
+
+def qubit_tapering_ref(tapered_qubits, ref):
+
+    """
+    This function removes the designated qubits from a given reference.
+
+    Arguments
+
+    tapered_qubits: list of integers
+        The ID numbers of the qubits to be tapered off.
+
+    ref: list of integers
+        A bit string representing the reference Slater determinant.
+    """
+
+    for i in tapered_qubits:
+        del ref[i]
+
+    return ref
--- a/src/qforte/utils/qbt_tapering.py
+++ b/src/qforte/utils/qbt_tapering.py
@@ -7,7 +7,6 @@ import numpy as np
 from itertools import product
 
 def find_Z2_symmetries(hamiltonian, qiskit_order = False):
-
     """
     This function computes the Z2 symmetries of the Hamiltonian, identifies the ID numbers of the qubits to
     be tapered off, provides the corresponding sign structures, and constructs the necessary unitary operators.
@@ -37,21 +36,21 @@ def find_Z2_symmetries(hamiltonian, qiskit_order = False):
     n_qubits = hamiltonian.num_qubits()
     n_strings = len(hamiltonian.terms())
 
-    ## Construct parity check matrix
+    ## Construct parity check matrix.
+    ## The rows of the parity check matrix correspond to Pauli strings of the qubit Hamiltonian.
+    ## The columns of the parity check matrix correspond to the Z_(n_qubits - 1), ..., Z_0, X_(n_qubits - 1), ..., X_0 Pauli gates.
+    ## For example, prt_chck_mtrx[i, n_qubits - 1] == 1 (0) designates the presence (absence) of the Z_0 Pauli gate in the ith Pauli string of the Hamiltonian.
     prt_chck_mtrx = np.zeros((n_strings, 2 * n_qubits), dtype=np.uint8)
-    for i, string in enumerate(hamiltonian.terms()):
-        for pauli in string[1].gates():
+    for i, (_, string) in enumerate(hamiltonian.terms()):
+        for pauli in string.gates():
             XYZ_gate = pauli.gate_id()
             trgt = pauli.target()
-            if XYZ_gate == 'Z':
+            if XYZ_gate in ['Y', 'Z']:
                 prt_chck_mtrx[i, n_qubits - trgt - 1] = 1
-            elif XYZ_gate == 'X':
-                prt_chck_mtrx[i, 2 * n_qubits - trgt - 1] = 1
-            else:
-                prt_chck_mtrx[i, n_qubits - trgt - 1] = 1
+            if XYZ_gate in ['X', 'Y']:
                 prt_chck_mtrx[i, 2 * n_qubits - trgt - 1] = 1
 
-    # 2) Find the basis of the kernel of the parity check matrix, ker{E}
+    # 2) Find a basis of the kernel of the parity check matrix
     # See https://en.wikipedia.org/wiki/Kernel_(linear_algebra)#Computation_by_Gaussian_elimination
 
     ## Augment the (n_strings) × (2 * n_qubits) parity check matrix by the (2 * n_qubits) × (2 * n_qubits) identity matrix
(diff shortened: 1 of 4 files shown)


src/qforte/utils/qbt_tapering.py
```diff
@@ -0,0 +1,318 @@
...
+    # 5) Compute the various signs corresponding to the eigenvalues of the sigma_x operators that will be removed
+
+    # 1) Convert qubit Hamiltonian to binary matrix
+
+    ## Obtain number of qubits and number of Pauli strings contained in qubit Hamiltonian
+    n_qubits = hamiltonian.num_qubits()
+    n_strings = len(hamiltonian.terms())
+
+    ## Construct parity check matrix
+    prt_chck_mtrx = np.zeros((n_strings, 2 * n_qubits), dtype=np.uint8)
```
JonathonMisiewicz, line 41: You should explicitly state the indexing conventions. Rows are Hamiltonian strings, which is simple enough. What is less straightforward is that qubits are indexed z_(n+1), ... z_0, x_(n+1), ..., x_0 rather than z_0, ..., z_(n+1), x_0, ..., x_(n+1)

src/qforte/utils/qbt_tapering.py
```diff
@@ -0,0 +1,318 @@
...
+    prt_chck_mtrx_aug = np.concatenate((prt_chck_mtrx, np.identity(2 * n_qubits, dtype=np.uint8)), axis=0)
+
+    ## Find column echelon form (CEF) of augmented matix.
+
+    idx = 0
+    for row in range(n_strings):
+        ### Find a non-zero row
+        if np.any(prt_chck_mtrx_aug[row,idx:] == 1):
+            ### Find positions of non-zero elements
+            count = np.argwhere(prt_chck_mtrx_aug[row,idx:] == 1)
```
JonathonMisiewicz, line 68: Why not move the `argwhere` up? Instead of the any check, you can check if the result of `argwhere` is empty or not.

Similarly, please rename the variable. This is not a count.

JonathonMisiewicz (changes requested):
I've finished an initial pass, but I'll need to go over this again. Please use descriptive variable names. The technical sub-algorithms of this PR are difficult to read due to variables not quite meaning what their names say they mean.

src/qforte/utils/qbt_tapering.py
```diff
@@ -0,0 +1,318 @@
...
+    gnrtrs = []
+    for i in range(tau.shape[0]):
+        pauli = qf.Circuit()
+        for j in reversed(range(n_qubits)):
+            if tau[i, j] == 1 and tau[i, j + n_qubits] == 1:
+                pauli.add_gate(qf.gate('Y', n_qubits - j - 1))
+            elif tau[i, j] == 1:
+                pauli.add_gate(qf.gate('X', n_qubits - j - 1))
+            elif tau[i, j + n_qubits] == 1:
+                pauli.add_gate(qf.gate('Z', n_qubits - j - 1))
```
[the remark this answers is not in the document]
imagoulas: I had to make a slight modification to your suggestion to avoid issues with the identity operator.

src/qforte/utils/qbt_tapering.py
```diff
@@ -0,0 +1,318 @@
...
+    for op in untrs[1:]:
+        unitary.operator_product(op, True, True)
+
+    # 5) Compute the various signs corresponding to the eigenvalues of the sigma_x operators that will be removed
+
+    signs = list(product([1, -1], repeat=len(sigma_x)))
+
+    return gnrtrs, sigma_x, untrs, unitary, signs
+
+def qubit_tapering_oprtr(n_qubits, tapered_qubits, sign, operator, unitary):
```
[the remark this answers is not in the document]
imagoulas: Good catch. It was used in an earlier implementation, not anymore.

src/qforte/utils/qbt_tapering.py
imagoulas: I understand that I can replace the `any` check by checking if the result of `argwhere` is empty or not. I am wondering if there is a computational benefit for doing so. Stylistically, I would prefer the `any` check. It seems more natural to me.

I named the variable `count` because `count.shape[0]` is the number of non-zero elements in a given row. I could rename `count` to `pos` since, technically, it holds the positions of the non-zero elements in a given row. What do you think?

JonathonMisiewicz: On further examination, `prt_chck_mtrx_aug[row,idx:] == 1` is just `prt_chck_mtrx_aug[row,idx:]`. New suggestion: Save that to a new variable and pass that to the two places you call `prt_chck_mtrx_aug[row,idx:] == 1`. My concern that motivated moving `argwhere` up was the duplication of that `prt_chck_matrix_aug` call. The new proposal also eliminates that.

While `pos` is technically correct, it's even less descriptive. Name it something like `nonzero_indices`.

imagoulas: Good catch regarding `prt_chck_mtrx_aug[row,idx:] == 1`. Could you please explain the benefit of storing `prt_chck_mtrx_aug[row,idx:]` into a new variable instead of passing it twice as is? The reason I am asking is that if we decide to make this change, I will need to make similar modifications to many places in the code.

JonathonMisiewicz: I don't expect a significant performance impact; this is pure readability.

imagoulas: In that case, do you mind if I just pass `prt_chck_mtrx_aug[row,idx:]` and similar for all other occurrences? I am afraid that creating new variables might be confusing when someone reads the code for the first time.

JonathonMisiewicz: As somebody who has been reading the code for the first time, the current approach is more confusing. I'm still not clear on what `prt_chck_mtrx_aug[row,idx:]` _means_.

imagoulas: Thank you for your input.

I know that most of what I will right is already known to you, but for the sake of completeness I will still write it.

In that part of the code, I am finding the column echelon form of the augmented parity check matrix (`prt_chck_mtrx_aug`). This is done by looping over the rows (`row`) of the matrix. The variable `idx` holds the index of the next column to be brought into column echelon form. Essentially, this implementation splits the parity check matrix into two submatrices. The first one is already in column echelon form. The second one is not. The variable `idx` marks the first column of the submatrix that still needs to be brought to echelon form.

My problem is how to come up with meaningful variable names for all places that need to be modified.

JonathonMisiewicz: Are you telling me that's your problem or are you asking me to propose better variable names?

imagoulas: Why not both :-)?

In any case, I am all for making the code more readable, but, as I already stated, I find the idea of adding more variables confusing, at least to me. If you have some suggestions, feel free to share them with me. We should keep in mind that the code is not supposed to be a manual, in the sense that I am not expecting people to read the code and _learn_ how qubit tapering _is_ performed. Apart from the original paper, which is also cited in the comments, we have the notes and presentation for it.

I hope that, based on my previous explanation, _prt_chck_mtrx_aug[row,idx:]_ is clear.

JonathonMisiewicz: I do not understand what could be confusing about creating a variable and then using that variable on each of the next two lines of code.

It looks like you need me to take a more active role in this review than I was expecting. I can do that, but I have a _lot_ of other things on my to-do list right now, so expect my response times to drop. Just push what you have to GitHub, and I'll get to it when I need a break from the rest of my list.

imagoulas: Let me start by saying that `I already created` a new variable and further documented the code based on my explanation of `prt_chck_mtrx_aug[row,idx:]` posted above. Hopefully this new explanation will clear any confusion.

I apologize if my replies came across as if "I need you to take a more active role in this review". That was not and still is not my intention. The reason I was open to suggestions for an extra variable name is that the code _makes perfect sense to me_ the way it is. My goal is to make the code more readable to `you`, since you represent someone who is reading the code for the first time.

Nevertheless, let me state my case. In my mind, an additional variable is redundant and, thus, confusing to `me`, since there is already a variable with a quite descriptive name, `prt_chck_mtrx_aug`. In that instance of the code, I am focusing on a slice of a given `row`, hence `prt_chck_mtrx_aug[row,idx:]`. I understand if one finds the name `idx` cryptic and confusing. It would make more sense to me to change the name of the variable `idx` to something more descriptive and meaningful than to store `prt_chck_mtrx_aug[row,idx:]` in a new variable. An additional reason that I do not like creating an unnecessary variable is that someone that reads the code will be forced to search where the variable is defined. I do understand that this is not an issue here, because the variable will be defined just a couple of lines above its usage, but still I find it bad practice. In principle an extra variable will also add to the memory cost. Again, I do understand that in this particular example the additional memory requirement is negligible.

I would say let us put an end to this exchange of messages on github, because it is quite time consuming and we both need to focus on more important things than a single variable, although it is nice that we are both passionate about it. As I said in the beginning, I already introduced a new variable, so we should be good to go. I will be pushing a commit soon.

JonathonMisiewicz:
Please add a test case that failed before your bug fix and that now passes.

src/qforte/utils/qbt_tapering.py
```diff
@@ -0,0 +1,321 @@
...
+    operator: QubitOperator
+        The operator that we wish to transform. For example, the qubit Hamiltonian or the UCC ansatz.
+
+    unitary: QubitOperator
+        The unitary operator that will perform the dersired transformation.
+    """
+
+    # Similarity transform the given operator using the fact that the unitary matrix is also Hermitian
+
+    ## This is necessary; otherwise later on we would be modifying the "unitary" object directly, which we do not want to do
```
JonathonMisiewicz, line 268: These comments are verbose.
`We copy these operators to prevent changing the original arguments.` is all you need to say: you don't need to say it here and then say it again for `operator_tmp`.

src/qforte/utils/qbt_tapering.py
```diff
@@ -0,0 +1,321 @@
...
+    unitary_tmp = qf.QubitOperator()
+    unitary_tmp.add_op(unitary)
+
+    ## This is necessary; if we do not do this, when transforming the Hamiltonian we would be changing the "hamiltonian"
+    ## attribute of the Molecule class to nonsense.
+    operator_tmp = qf.QubitOperator()
+    operator_tmp.add_op(operator)
```
JonathonMisiewicz, line 275: This is ugly. We really should define a copy function C-side. Beyond the scope of this PR, sadly, but we should file an issue.

src/qforte/utils/qbt_tapering.py
```diff
@@ -0,0 +1,321 @@
...
+
+    operator_tmp.operator_product(unitary_tmp, True, True)
+    unitary_tmp.operator_product(operator_tmp, True, True)
+    operator_tmp = unitary_tmp
+
+    # Create a QForte QubitOperator object that will store the tapered operator
+    operator_tapered = qf.QubitOperator()
+
+    for i, (coeff, circuit) in enumerate(operator_tmp.terms()):
```
JonathonMisiewicz, line 284: This is _also_ ugly. Why do we not define `__iter__` for `QubitOperator`?

src/qforte/utils/qbt_tapering.py
```diff
@@ -0,0 +1,321 @@
...
+        # Create circuit that will store the pauli string with gates corresponding to qubits to be tapered off removed
+        circ = qf.Circuit()
+        # List that will store the coefficients modified according to the sign structure provided by the user
+        for pauli in circuit.gates():
```
JonathonMisiewicz, line 288: Ugly again. We should define an `__iter__` for a circuit. Again, separate issue.

src/qforte/utils/qbt_tapering.py
```diff
@@ -0,0 +1,321 @@
...
+    operator_tmp.operator_product(unitary_tmp, True, True)
+    unitary_tmp.operator_product(operator_tmp, True, True)
+    operator_tmp = unitary_tmp
+
+    # Create a QForte QubitOperator object that will store the tapered operator
+    operator_tapered = qf.QubitOperator()
+
+    for i, (coeff, circuit) in enumerate(operator_tmp.terms()):
+        # Create circuit that will store the pauli string with gates corresponding to qubits to be tapered off removed
+        circ = qf.Circuit()
```
JonathonMisiewicz, line 286: Rename this to `tapered_circuit`. With the improved name, the above comment is redundant and should be removed.

src/qforte/utils/qbt_tapering.py
```diff
@@ -0,0 +1,321 @@
...
+        # List that will store the coefficients modified according to the sign structure provided by the user
+        for pauli in circuit.gates():
+            if any(qubit == pauli.target() for qubit in tapered_qubits):
+                coeff *= sign[int(np.argwhere(tapered_qubits == pauli.target()))]
+            else:
+                # Adjust the qubit ID of a given gate based on the number of gates that were removed before it
```
JonathonMisiewicz, line 292: Change the comment to : "adjust gate targets to account for eliminated operators". This is shorter and more precise.

src/qforte/utils/qbt_tapering.py
```diff
@@ -0,0 +1,321 @@
...
+                idx = np.count_nonzero(pauli.target() > tapered_qubits)
+                target = pauli.target() - idx
+                gate = pauli.gate_id()
+                circ.add_gate(qf.gate(gate, target))
+
+        operator_tapered.add_term(coeff, circ)
+
+    # Simplify the entire operator
```
JonathonMisiewicz, line 300:
```suggestion
```

This comment adds nothing. It's clear that calling `simplify` is intended to simplify the operator.

src/qforte/utils/qbt_tapering.py
```diff
@@ -0,0 +1,321 @@
...
+    a single sigma_x operator. The sigma_x gates acting on qubits to be tapered off are replaced by their
+    +/- 1 eigenvalues, which are provided by the user.
+
+    Arguments
+
+    tapered_qubits: list of integers
+        The ID numbers of the qubits to be tapered off.
+
+    sign: list of +/- 1.
+        The desired sign structure corresponding to the eigenvalues of the sigma_x operators acting on
```
JonathonMisiewicz, line 256: "sign structure corresponding" is vague.

I think you mean _For each qubit to be tapered, should we assume we are in the + or - eigenspace? Takes one argument for each._

src/qforte/utils/qbt_tapering.py
```diff
@@ -0,0 +1,321 @@
...
+    Arguments:
+
+    hamiltonian: QubitOperator
+        The qubit Hamiltonian whose Z2 symmetries we want to find.
+
+    qiskit_order: bool
+        If True, the qubits to be tapered off will have the smallest id numbers possible, as is done in Qiskit.
+        If False, the qubits to be tapered off will have the largest id numbers possible. This is the default,
+        since it leads to less operations once the qubits are removed.
+    """
```
JonathonMisiewicz, line 24: Explicitly specify the return values.

src/qforte/utils/qbt_tapering.py
```diff
@@ -0,0 +1,321 @@
...
+
+    # The algorithm consists of the following steps:
+    # 1) Convert qubit Hamiltonian to binary matrix, called the parity check matrix E
+    # 2) Find the basis of the kernel of the parity check matrix, ker{E}
+    # 3) Find maximal Abelian subgroup of the basis of ker{E}
+    # 4) Construct unitary matrices that will be used in the qubit_tapering algorithm
+    # 5) Compute the various signs corresponding to the eigenvalues of the sigma_x operators that will be removed
+
+    # 1) Convert qubit Hamiltonian to binary matrix
```
JonathonMisiewicz, line 33: You are not converting a Hamiltonian to the party check matrix. You are constructing the parity check matrix from a Hamiltonian.

src/qforte/utils/qbt_tapering.py
```diff
@@ -0,0 +1,321 @@
...
+
+    ## Obtain number of qubits and number of Pauli strings contained in qubit Hamiltonian
```
JonathonMisiewicz, line 35: Completely redundant comment.
```suggestion
```

src/qforte/utils/qbt_tapering.py
```diff
@@ -0,0 +1,321 @@
...
+    This function computes the Z2 symmetries of the Hamiltonian, identifies the ID numbers of the qubits to
+    be tapered off, provides the corresponding sign structures, and constructs the necessary unitary operators.
+    It is based on the qubit tapering approach reported in https://arxiv.org/abs/1701.08213.
+
+    Arguments:
+
+    hamiltonian: QubitOperator
+        The qubit Hamiltonian whose Z2 symmetries we want to find.
+
+    qiskit_order: bool
```
JonathonMisiewicz, line 20: I think it would be better to rename this variable to `taper_from_least`.

src/qforte/utils/qbt_tapering.py
```diff
@@ -0,0 +1,321 @@
...
+    # 4) Construct unitary matrices that will be used in the qubit_tapering algorithm
+    # 5) Compute the various signs corresponding to the eigenvalues of the sigma_x operators that will be removed
+
+    # 1) Convert qubit Hamiltonian to binary matrix
+
+    ## Obtain number of qubits and number of Pauli strings contained in qubit Hamiltonian
+    n_qubits = hamiltonian.num_qubits()
+    n_strings = len(hamiltonian.terms())
+
+    ## Construct parity check matrix.
```
JonathonMisiewicz, line 39:
```suggestion
```

Redundant comment.

src/qforte/utils/qbt_tapering.py
```diff
@@ -0,0 +1,321 @@
...
+    ## The rows of the parity check matrix correspond to Pauli strings of the qubit Hamiltonian.
+    ## The columns of the parity check matrix correspond to the Z_(n_qubits - 1), ..., Z_0, X_(n_qubits - 1), ..., X_0 Pauli gates.
+    ## For example, prt_chck_mtrx[i, n_qubits - 1] == 1 (0) designates the presence (absence) of the Z_0 Pauli gate in the ith Pauli string of the Hamiltonian.
```
JonathonMisiewicz, line 42:
```suggestion
```

src/qforte/utils/qbt_tapering.py
```diff
@@ -0,0 +1,321 @@
...
+    # 2) Find a basis of the kernel of the parity check matrix
+    # See https://en.wikipedia.org/wiki/Kernel_(linear_algebra)#Computation_by_Gaussian_elimination
+
+    ## Augment the (n_strings) × (2 * n_qubits) parity check matrix by the (2 * n_qubits) × (2 * n_qubits) identity matrix
+
+    prt_chck_mtrx_aug = np.concatenate((prt_chck_mtrx, np.identity(2 * n_qubits, dtype=np.uint8)), axis=0)
+
+    ## Find column echelon form (CEF) of augmented matix.
+
+    ### To speed up the process, we only work with the part of the parity check matrix that has not been transformed yet to CEF.
```
JonathonMisiewicz, line 62:
```suggestion
    ### We only work with the column of the parity check matrix that have not been transformed yet to CEF.
```

src/qforte/utils/qbt_tapering.py
```diff
@@ -0,0 +1,321 @@
...
+    ### This is controlled by the "idx" variable
+    idx = 0
+
+    for row in range(n_strings):
+        ### Find a non-zero row
+        clmns_not_in_CEF = prt_chck_mtrx_aug[row,idx:]
+        if np.any(clmns_not_in_CEF):
+            ### Find positions of non-zero elements
```
JonathonMisiewicz, line 70:
```suggestion
```

Redundant thanks to variable renaming.

src/qforte/utils/qbt_tapering.py
```diff
@@ -0,0 +1,321 @@
...
+    prt_chck_mtrx_aug = np.concatenate((prt_chck_mtrx, np.identity(2 * n_qubits, dtype=np.uint8)), axis=0)
+
+    ## Find column echelon form (CEF) of augmented matix.
+
+    ### To speed up the process, we only work with the part of the parity check matrix that has not been transformed yet to CEF.
+    ### This is controlled by the "idx" variable
+    idx = 0
+
+    for row in range(n_strings):
+        ### Find a non-zero row
```
JonathonMisiewicz, line 67: This comment seems inaccurate. Do you mean "determine if the column is not yet in CEF"?

src/qforte/utils/qbt_tapering.py
```diff
@@ -0,0 +1,321 @@
...
+        clmns_not_in_CEF = prt_chck_mtrx_aug[row,idx:]
+        if np.any(clmns_not_in_CEF):
+            ### Find positions of non-zero elements
+            nonzero_indices = np.argwhere(clmns_not_in_CEF)
+            column = int(nonzero_indices[0]) + idx
+            ### Swap columns to bring first non-zero element in this row immediately to the right of that of the previous row
```
JonathonMisiewicz, line 73:
```suggestion
            ### Bring first non-zero element in this row immediately to the right of that of the previous row
```

src/qforte/utils/qbt_tapering.py
```diff
@@ -0,0 +1,321 @@
...
+    ### This is controlled by the "idx" variable
+    idx = 0
+
+    for row in range(n_strings):
+        ### Find a non-zero row
+        clmns_not_in_CEF = prt_chck_mtrx_aug[row,idx:]
+        if np.any(clmns_not_in_CEF):
+            ### Find positions of non-zero elements
+            nonzero_indices = np.argwhere(clmns_not_in_CEF)
+            column = int(nonzero_indices[0]) + idx
```
JonathonMisiewicz, line 72: Rename variable to something more descriptive. There are multiple columns: what's special about this one?

src/qforte/utils/qbt_tapering.py
```diff
@@ -0,0 +1,321 @@
...
+            ### Swap columns to bring first non-zero element in this row immediately to the right of that of the previous row
+            prt_chck_mtrx_aug[:, [column , idx]] = prt_chck_mtrx_aug[:, [idx, column]]
+            ### Eliminate remaining non-zero elements in the submatrix
```
JonathonMisiewicz, line 75: This is the first time you're introduced the idea of a submatrix. Please describe in more detail.

src/qforte/utils/qbt_tapering.py
```diff
@@ -0,0 +1,321 @@
...
+            for i in range(1,nonzero_indices.shape[0]):
+                column2 = int(nonzero_indices[i]) + idx
+                prt_chck_mtrx_aug[:,column2] = np.bitwise_xor(prt_chck_mtrx_aug[:, idx], prt_chck_mtrx_aug[:,column2])
+            ### decrease the number of columns that need to be checked
+            idx += 1
+
+    ## Find basis vectors of kernel by checking for zero columns in the CEF of the parity check matrix
+
+    dim = 0
```
JonathonMisiewicz, line 84: Please use a `bool` rather than an `int`.

src/qforte/utils/qbt_tapering.py
```diff
@@ -0,0 +1,321 @@
...
+    basis = np.zeros((1, 2 * n_qubits), dtype=np.uint8)
+    for column in range(2 * n_qubits):
+        row = n_strings
+        if not np.any(prt_chck_mtrx_aug[:row, column]):
+            if dim == 0:
+                basis = prt_chck_mtrx_aug[row:, column].reshape((1,2 * n_qubits))
```
JonathonMisiewicz, line 90: You need to compute `prt_chck_mtrx_aug[row:, column].reshape((1,2 * n_qubits))` in either case, so why not pull this out of the `if` statement?

src/qforte/utils/qbt_tapering.py
```diff
@@ -0,0 +1,321 @@
...
+            ### decrease the number of columns that need to be checked
+            idx += 1
+
+    ## Find basis vectors of kernel by checking for zero columns in the CEF of the parity check matrix
+
+    dim = 0
+    basis = np.zeros((1, 2 * n_qubits), dtype=np.uint8)
+    for column in range(2 * n_qubits):
+        row = n_strings
+        if not np.any(prt_chck_mtrx_aug[:row, column]):
```
JonathonMisiewicz, line 88: Are the indices for which this is true related to the variable `idx` from earlier?

src/qforte/utils/qbt_tapering.py
```diff
@@ -0,0 +1,321 @@
...
+    blnr_frm_mtrx = np.concatenate((top, bottom), axis=0)
+
+    ## Construct abln matrix.
+    ## abln[i,j] contains the value of the bilinear form betweeen the ith and jth basis vectors of ker{E}
+    ## A "0"/"1" in the ith row and jth column of abln means that the Pauli strings
+    ## associated with the ith and jth binary vectors commute/anticommute.
+
+    aux = np.matmul(blnr_frm_mtrx, basis.T)
+
+    abln = np.zeros((basis.shape[0], basis.shape[0]), dtype=np.uint8)
```
JonathonMisiewicz, line 113: Expand the variable name. I can understand cutting vowels from long variable names, but it's only four letters at present. Adding three to get `abelian` is more readable and not significantly more typing.

src/qforte/utils/qbt_tapering.py
```diff
@@ -0,0 +1,321 @@
...
+    # 3) Find maximal Abelian subgroup of the basis of ker{E}
+    # This is done using the symplectic bilinear form B that the binary vector space is equipped with:
+    # B(v,w) = v * B * w.T,
+    # where v and w are binary vectors and B is the matrix of the bilinear form.
+
+    ## Construct matrix of bilinear form.
+
+    top = np.concatenate((np.zeros((n_qubits, n_qubits), dtype=np.uint8), np.identity(n_qubits, dtype=np.uint8)), axis=1)
+    bottom = np.concatenate((np.identity(n_qubits, dtype=np.uint8), np.zeros((n_qubits, n_qubits), dtype=np.uint8)), axis=1)
+    blnr_frm_mtrx = np.concatenate((top, bottom), axis=0)
```
JonathonMisiewicz, line 104: Numpy's [block](https://numpy.org/doc/stable/reference/generated/numpy.block.html) should make what you're trying to do more transparent.

src/qforte/utils/qbt_tapering.py
```diff
@@ -0,0 +1,321 @@
...
+
+    ## Construct abln matrix.
+    ## abln[i,j] contains the value of the bilinear form betweeen the ith and jth basis vectors of ker{E}
+    ## A "0"/"1" in the ith row and jth column of abln means that the Pauli strings
```
JonathonMisiewicz, line 108:
```suggestion
    ## 0/1 in the ith row and jth column of abln means that the Pauli strings
```

src/qforte/utils/qbt_tapering.py
```diff
@@ -0,0 +1,321 @@
...
+    ## associated with the ith and jth binary vectors commute/anticommute.
+
+    aux = np.matmul(blnr_frm_mtrx, basis.T)
+
+    abln = np.zeros((basis.shape[0], basis.shape[0]), dtype=np.uint8)
+    for row in range(basis.shape[0]):
```
JonathonMisiewicz, line 114: This is overcomplicated. Just do matrix multiplication, and then apply the % to each element. You can probably just do `abln = abln % 2`.

src/qforte/utils/qbt_tapering.py
```diff
@@ -0,0 +1,321 @@
...
+            if dim == 0:
+                basis = prt_chck_mtrx_aug[row:, column].reshape((1,2 * n_qubits))
+                dim += 1
+            else:
+                basis = np.concatenate((basis, prt_chck_mtrx_aug[row:, column].reshape((1,2 * n_qubits))), axis=0)
+
+    # 3) Find maximal Abelian subgroup of the basis of ker{E}
+    # This is done using the symplectic bilinear form B that the binary vector space is equipped with:
+    # B(v,w) = v * B * w.T,
+    # where v and w are binary vectors and B is the matrix of the bilinear form.
```
JonathonMisiewicz, line 98: Proposal for a better way to write the comment: "This is done by constructing an intermediate B so that v * B * w.T = 0 means v and w commute, and v * B * w.T = 1 means they anticommute."

src/qforte/utils/qbt_tapering.py
```diff
@@ -0,0 +1,321 @@
...
+
+    aux = np.matmul(blnr_frm_mtrx, basis.T)
+
+    abln = np.zeros((basis.shape[0], basis.shape[0]), dtype=np.uint8)
+    for row in range(basis.shape[0]):
+        for column in range(basis.shape[0]):
+            for idx in range(2*n_qubits):
+                abln[row, column] = (abln[row, column] + basis[row, idx] * aux[idx, column]) % 2
+
+    ## Find the maximal abelian subgroup iteratively. This is done as follows:
```
JonathonMisiewicz, line 119: Where does this algorithm come from?

src/qforte/utils/qbt_tapering.py
```diff
@@ -0,0 +1,321 @@
...
+
+        ### Construct abln matrix for the next iteration.
+
+        aux = np.matmul(blnr_frm_mtrx, basis_tmp.T)
+
+        abln = np.zeros((basis_tmp.shape[0], basis_tmp.shape[0]), dtype=np.uint8)
+        for row in range(basis_tmp.shape[0]):
+            for column in range(basis_tmp.shape[0]):
+                for idx in range(2*n_qubits):
+                    abln[row, column] = (abln[row, column] + basis_tmp[row, idx] * aux[idx, column]) % 2
```
JonathonMisiewicz, line 153: Again, this is overcomplicated when you can do matrix multiplication.

src/qforte/utils/qbt_tapering.py
```diff
@@ -0,0 +1,321 @@
...
+
+    tau = basis_tmp
```
JonathonMisiewicz, line 155: What is tau? A more descriptive variable name may be in order.

src/qforte/utils/qbt_tapering.py
```diff
@@ -0,0 +1,321 @@
...
+
+    #### This step is not necessary. It transforms the generators of the symmetry group in such a way that the
```
JonathonMisiewicz, line 157: Lower the # count from 4 to 3.

src/qforte/utils/qbt_tapering.py
```diff
@@ -0,0 +1,321 @@
...
+    #### removed qubits have the smallest/largest possbile identity numbers. A similar modification is made
+    #### at the "sigma_x selection" part of the algorithm below. The "smallest" case produces identical resutls
+    #### with those obtained using Qiskit, after taking into account that in Qiskit Slater determinants are
+    #### represented by strings of alpha orbitals followed by the beta ones.
```
JonathonMisiewicz, line 161:
```suggestion
    #### at the "sigma_x selection" part of the algorithm below.
```

We don't want our comment to be invalidated, should qiskit ever change their code.

src/qforte/utils/qbt_tapering.py
JonathonMisiewicz, line 157: I'll _eventually_ want this part of the code to be more flexible, but that can wait.

My experimenting with TFIM suggests that on actual hardware, "which qubit do I taper" has surprising ramifications.

JonathonMisiewicz (changes requested):
I'll need to give this another pass.

The new variable names were very helpful in reading the code.

src/qforte/utils/qbt_tapering.py
```diff
@@ -0,0 +1,321 @@
...
+
+    tapered_qubits: list of integers
+        The ID numbers of the qubits to be tapered off.
+
+    sign: list of +/- 1.
+        The desired sign structure corresponding to the eigenvalues of the sigma_x operators acting on
+        the qubits to be tapered off.
+
+    operator: QubitOperator
+        The operator that we wish to transform. For example, the qubit Hamiltonian or the UCC ansatz.
```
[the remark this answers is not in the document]
imagoulas: Truer words have never been spoken. In my mind, I was equating the ansatz with the wave operator.

src/qforte/utils/qbt_tapering.py
imagoulas: I came up with this algorithm.

src/qforte/utils/qbt_tapering.py
```diff
@@ -0,0 +1,321 @@
...
+    After the unitary transformation, the resulting operator acts on the qubits to be tapered off by at most
+    a single sigma_x operator. The sigma_x gates acting on qubits to be tapered off are replaced by their
+    +/- 1 eigenvalues, which are provided by the user.
+
+    Arguments
+
+    tapered_qubits: list of integers
+        The ID numbers of the qubits to be tapered off.
+
+    sign: list of +/- 1.
```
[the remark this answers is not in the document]
imagoulas: Could you please explain what "validate this argument" means?

src/qforte/utils/qbt_tapering.py
imagoulas: Indeed, they are. According to the earlier part of the algorithm, the first `idx` columns of the CEFed parity check matrix are nonzero by construction. I modified line 86 above to reflect this fact.

src/qforte/utils/qubit_tapering.py
```diff
@@ -0,0 +1,345 @@
...
+            ### Eliminate remaining non-zero row elements in the part of the parity check matrix that is not in CEF
+            for i in range(1,nonzero_indices.shape[0]):
+                column_not_in_CEF = int(nonzero_indices[i]) + idx
+                prt_chck_mtrx_aug[:,column_not_in_CEF] = np.bitwise_xor(prt_chck_mtrx_aug[:, idx], prt_chck_mtrx_aug[:,column_not_in_CEF])
+            ### decrease the number of columns that need to be checked
+            idx += 1
+
+    ## Find basis vectors of kernel by checking for zero columns in the CEF of the parity check matrix
+
+    dim = False
```
JonathonMisiewicz, line 99: `dim` is not descriptive and you don't use the value of `basis` you initialize below. Can you initialize `basis` with shape `(0, 2 * n_qubits)`? Then you can always concatenate `new_basis_vector` to `basis` and don't need `dim`.

src/qforte/utils/qubit_tapering.py
```diff
@@ -0,0 +1,345 @@
...
+            else:
+                basis = np.concatenate((basis, new_basis_vector), axis=0)
+
+    # 3) Find maximal Abelian subgroup of the basis of ker{E}
+    # This is done using the symplectic bilinear form B that the binary vector space is equipped with:
+    # B(v,w) = v * B * w.T,
+    # where v and w are binary vectors and B is the matrix of the bilinear form.
+    # B(v,w) = 0 means that v and w commute while B(v,w) = 1 means that they anticommute.
+
+    ## Construct matrix of bilinear form.
```
JonathonMisiewicz, line 118:
```suggestion
    ## Construct matrix B, which acts on binary vectors. v * B * w.T = 0 means v and w commute.
    ## v * B * w.T = 1 means anticommute.
```

Original comment was verbose.

src/qforte/utils/qubit_tapering.py
```diff
@@ -0,0 +1,345 @@
...
+    identity_matrix = np.identity(n_qubits, dtype=np.uint8)
+    blnr_frm_mtrx = np.block([
+        [zero_matrix    , identity_matrix],
+        [identity_matrix, zero_matrix    ]
+        ])
+
+    ## Construct abelian matrix.
+    ## abelian[i,j] contains the value of the bilinear form betweeen the ith and jth basis vectors of ker{E}
+    ## 0/1 in the ith row and jth column of abelian means that the Pauli strings
+    ## associated with the ith and jth binary vectors commute/anticommute.
```
JonathonMisiewicz, line 130:
```suggestion
    ## For all pairs of ker(E) basis vectors, compute abelian[v, w] = v * B * w.T
```
Original comment was verbose.

src/qforte/utils/qubit_tapering.py
```diff
@@ -0,0 +1,345 @@
...
+
+    abelian = np.matmul(basis, np.matmul(blnr_frm_mtrx, basis.T))
+    abelian %= 2
+
+    ## Find the maximal abelian subgroup iteratively. This is done as follows:
```
JonathonMisiewicz, line 135: I'm not convinced that this is going to get the maximal abelian subgroup. How much looking have you done into algorithms for this? A quick Google search gives me one [alternative](https://math.stackexchange.com/questions/473502/how-to-find-maximal-abelian-subgroups/473921), but I haven't done much thinking about it.

src/qforte/utils/qubit_tapering.py
```diff
@@ -0,0 +1,345 @@
...
+        ### Construct abelian matrix for the next iteration.
+
+        abelian = np.matmul(basis_tmp, np.matmul(blnr_frm_mtrx, basis_tmp.T))
+        abelian %= 2
+
+    generators_binary = basis_tmp
+
+    ### This step is not necessary. It transforms the generators of the symmetry group in such a way that the
+    ### removed qubits have the smallest/largest possbile identity numbers. A similar modification is made
+    ### at the "sigma_x selection" part of the algorithm below.
```
JonathonMisiewicz, line 170:
```suggestion
    ## Transform the generators to control which qubit we remove.
```

Original comment was verbose.

src/qforte/utils/qubit_tapering.py
```diff
@@ -0,0 +1,345 @@
...
+    idx = 0
+    lst = []
+    if taper_from_least:
+        _range = list(reversed(range(n_qubits, 2 * n_qubits)))
+    else:
+        _range = list(range(n_qubits, 2 * n_qubits))
```
JonathonMisiewicz, line 176: Why do you need to cast this to a `list`? If you're just iterating over `_range`, it should be unnecessary.

src/qforte/utils/qubit_tapering.py
```diff
@@ -0,0 +1,345 @@
...
+    for column in _range:
```
JonathonMisiewicz, line 177: I don't understand this part of your algorithm. Why are you modifying the _generators_? According to your presentation, to control which qubits you taper off, you choose the `x` gates you add in the unitary. What am I missing?

src/qforte/utils/qubit_tapering.py
JonathonMisiewicz: I should add that we _could_ save this for a later PR.

src/qforte/utils/qubit_tapering.py
```diff
@@ -0,0 +1,345 @@
...
+
+    generators: list of QubitOperator objects
+        Qubit representation of set of independent generators of the symmetry group of the Hamiltonian.
+
+    sigma_x: list of integers
+        The indices of the sigma_x Pauli gates that each generator will be transformed to.
+        The indices coincide with the indices of the qubits to be tapered off.
+
+    unitaries: list of QubitOperator objects
+        The Clifford gates that transform each generator to a given sigma_x Pauli gate.
```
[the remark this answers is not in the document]
imagoulas: The `generators` and `unitaries` are not needed by the other tapering routines and we do not have to return them. The only reason that I am returning these quantities is to be able to check against Qiskit.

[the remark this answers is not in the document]
JonathonMisiewicz: Replacing `gates` with `operators` sounds better to me: in the context of QForte, `gates` always means a gate you can physically implement. I have no strong opinion on whether we say `Clifford` or `unitary`.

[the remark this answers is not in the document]
JonathonMisiewicz: Can you give me a link to the counterpart function in Qiskit? I'd like to see how much their function returns.

src/qforte/utils/qubit_tapering.py
JonathonMisiewicz: What does Qiskit do to find the maximal abelian subgroup?

src/qforte/utils/qubit_tapering.py
```diff
@@ -7,7 +7,7 @@
 from itertools import product
 from copy import deepcopy
 
-def find_Z2_symmetries(hamiltonian, taper_from_least = False, debug = False):
+def find_Z2_symmetries(hamiltonian, taper_from_least = True, debug = False):
```
JonathonMisiewicz, line 10: Did you mean to change this flag?

src/qforte/utils/qubit_tapering.py
```diff
@@ -198,7 +202,91 @@ def find_parity_check_matrix_kernel(n_qubits, n_strings, prt_chck_mtrx):
...
+                    if pauli_2 not in processed:
+                        processed.append(pauli_2)
+                        for pauli in range(SGSO_basis.shape[0]):
+                            if pauli not in processed:
+                                if SGSO_commute[pauli, pauli_1] and SGSO_commute[pauli, pauli_2]:
+                                    SGSO_basis[pauli] ^= SGSO_basis[pauli_1] ^ SGSO_basis[pauli_2]
+                                elif SGSO_commute[pauli, pauli_1] and not SGSO_commute[pauli, pauli_2]:
+                                    SGSO_basis[pauli] ^= SGSO_basis[pauli_2]
+                                elif not SGSO_commute[pauli, pauli_1] and SGSO_commute[pauli, pauli_2]:
+                                    SGSO_basis[pauli] ^= SGSO_basis[pauli_1]
```
JonathonMisiewicz, line 284: It looks like you can rewrite this as
```
if SGSO_commute[pauli, pauli_2]:
    SGSO_basis[pauli] ^= SGSO_basis[pauli_1]
if SGSO_commute[pauli, pauli_1]:
    SGSO_basis[pauli] ^= SGSO_basis[pauli_2]
```
This seems simpler to me.

JonathonMisiewicz (approved):
I have two last comments, but feel free to merge once those are addressed.

src/qforte/utils/qubit_tapering.py
imagoulas: I did. Based on a few UCCSD runs (using my local version of QForte in which qubit tapering is interfaced with UCCNVQE), this option leads to quantum circuits with less CNOT gates (of course, this could always be a coincidence).

src/qforte/utils/qubit_tapering.py
imagoulas: This is indeed very compact!

JonathonMisiewicz:
Go ahead and merge when you're satisfied.

PR #156, opened 2021-09-27
Move UCC to Qubit Space

### LINKED ISSUE #127: Replace `_pool` with `_qubit_pool`
At present, the Algorithm class uses a `_pool_` attribute, which represents a linear combination of fermionic second quantized operators. This is inappropriate for systems that are not fermions, or for algorithms with an operator pool for which expressing things in terms of second quantization is inconvenient. These should be replaced with an operator pool expressed solely in terms of qubit-space operators.

JonathonMisiewicz:
## Description
Closes #127. `AnsatzAlgorithm._pool_obj` is banished to oblivion in favor of `AnastzAlgorithm._qubit_pool`.

This is major enough interest that I'm notifying @fevangelista. Review not required, but would be appreciated.

## User Notes
- [x] The `_pool_obj` is no longer on the `AnsatzAlgorithm` and has been replaced with `_qubit_pool`
- [x] Pybinding a few features of the `QubitOpPool`
- [x] `QubitOpPool` now has a description field, if you want to specify what these generators mean in some other space, i.e., fermion space
- [x]  `QubitOpPool::get_qubit_operator` added. This is rough and will need to be re-analyzer later.

## Checklist
- [x] Docstrings updated, where appropriate
- [x] Ready to go!

### DIFF
--- a/src/qforte/abc/ansatz.py
+++ b/src/qforte/abc/ansatz.py
@@ -26,10 +26,10 @@ class UCC:
             A list of parameters that define the variational degrees of freedom in
             the state preparation circuit Uvqc. This is needed for the scipy minimizer.
         """
-        temp_pool = qf.SQOpPool()
+        temp_pool = qf.QubitOpPool()
         tamps = self._tamps if amplitudes is None else amplitudes
         for tamp, top in zip(tamps, self._tops):
-            temp_pool.add(tamp, self._pool_obj[top][1])
+            temp_pool.add(tamp, self._qubit_pool[top][1])
 
         A = temp_pool.get_qubit_operator('commuting_grp_lex')
 
@@ -74,21 +74,25 @@ class UCC:
             residuals.
         """
 
+        reference_state = qf.QubitBasis(self._nqb)
+        for k, occ in enumerate(self._ref):
+            reference_state.set_bit(k, occ)
+
         resids_over_denoms = []
 
-        # loop over toperators
-        for mu, m in enumerate(self._tops):
-            sq_op = self._pool_obj[m][1]
+        for mu, (index, _) in enumerate(self._excited_dets):
+            excited = qf.QubitBasis(index)
 
-            temp_idx = sq_op.terms()[0][2][-1]
-            if temp_idx < int(sum(self._ref)/2): # if temp_idx is an occupied idx
-                sq_creators = sq_op.terms()[0][1]
-                sq_annihilators = sq_op.terms()[0][2]
-            else:
-                sq_creators = sq_op.terms()[0][2]
-                sq_annihilators = sq_op.terms()[0][1]
+            denom = 0
 
-            denom = sum(self._orb_e[x] for x in sq_annihilators) - sum(self._orb_e[x] for x in sq_creators)
+            for qubit in range(self._nqb):
+                if reference_state.get_bit(qubit):
+                    if not excited.get_bit(qubit):
+                        # Annnihilated
+                        denom += self._orb_e[qubit]
+                elif excited.get_bit(qubit):
+                    # Created
+                    denom -= self._orb_e[qubit]
 
             res_mu = residuals[mu] / denom
 
--- a/src/qforte/abc/ansatz.py
+++ b/src/qforte/abc/ansatz.py
@@ -64,7 +64,7 @@ class UCC:
             self._orb_e.append(ei)
 
     def get_res_over_mpdenom(self, residuals):
-        """This function returns a vector given by the residuals dividied by the
+        """This function returns a vector given by the residuals divided by the
         respective Moller Plesset denominators.
 
         Parameters
--- a/src/qforte/computer.h
+++ b/src/qforte/computer.h
@@ -73,7 +73,7 @@ class Computer {
     std::vector<std::complex<double>> direct_idxd_oppl_exp_val(const QubitOpPool& qopl, const std::vector<int>& idxs);
 
     /// measure expectation value of all operators in an operator pool, where
-    /// the operator coefficents have been multipiled by mults
+    /// the operator coefficients have been multiplied by mults
     std::vector<std::complex<double>> direct_oppl_exp_val_w_mults(
         const QubitOpPool& qopl,
         const std::vector<std::complex<double>>& mults);
--- a/src/qforte/computer.h
+++ b/src/qforte/computer.h
@@ -49,7 +49,7 @@ class Computer {
     void apply_gate(const Gate& qg);
 
     /// apply a constant to the quantum computer (WARNING, this operation
-    /// is not physical as it does not represent a unitary opperation). Only
+    /// is not physical as it does not represent a unitary operation). Only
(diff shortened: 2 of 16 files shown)


src/qforte/abc/ansatz.py
```diff
@@ -74,21 +74,25 @@ def get_res_over_mpdenom(self, residuals):
...
 
-            denom = sum(self._orb_e[x] for x in sq_annihilators) - sum(self._orb_e[x] for x in sq_creators)
+            for qubit in range(self._nqb):
+                if reference_state.get_bit(qubit):
+                    if not excited.get_bit(qubit):
+                        # Annnihilated
+                        denom += self._orb_e[qubit]
+                elif excited.get_bit(qubit):
+                    # Created
+                    denom -= self._orb_e[qubit]
```
fevangelista, line 96: I have a question here: do we assume a JW transformation here? In the old version based on the sq_operators, the denominator are independent of the fermionic mapping. But if we examine qubits it should depend if we use parity or JW. So if this is the case, I would consider adding a function that computes the denominator and that keeps into account the transformation used. We might have to think about this carefully to make it work in the general case.

src/qforte/computer.h
```diff
@@ -72,8 +72,8 @@ class Computer {
     /// measure expectation value for specific operators in an operator pool
     std::vector<std::complex<double>> direct_idxd_oppl_exp_val(const QubitOpPool& qopl, const std::vector<int>& idxs);
 
-    /// measure expectaion value of all operators in an operator pool, where the
-    /// operator coefficents have been multipild by mults
+    /// measure expectation value of all operators in an operator pool, where
+    /// the operator coefficents have been multipiled by mults
```
fevangelista, line 76:
```suggestion
    /// the operator coefficents have been multiplied by mults
```

src/qforte/qubit_op_pool.cc
```diff
@@ -43,6 +44,36 @@ const std::vector<std::pair<std::complex<double>, QubitOperator>>& QubitOpPool::
...
+    } else if (order_type=="commuting_grp_lex") {
+        for (auto& term : terms_) {
+            auto child = term.second;
+            child.mult_coeffs(term.first);
+            child.simplify(combine_like_terms=combine_like_terms);
+            child.order_terms();
+            parent.add_op(child);
+        }
+    } else {
+        throw std::invalid_argument( "Invalid order_type specified.");
```
fevangelista, line 72:
```suggestion
        throw std::invalid_argument( "Invalid order_type specified. Allowed values are unique_lex and commuting_grp_lex");
```

src/qforte/qubit_op_pool.h
```diff
@@ -21,7 +21,7 @@ class QubitOpPool {
     void set_terms(std::vector<std::pair<std::complex<double>, QubitOperator>>& new_terms);
 
     /// add one set of anihilators and/or creators to the second quantized operator pool
```
fevangelista, line 23: Is this documentation string accurate?

fevangelista (approved):
Thank you. This refactoring makes a lot of sense. I left a few suggestions for improvements and comments regarding potential future issues with other fermionic mapping.

src/qforte/abc/ansatz.py
JonathonMisiewicz: Strictly speaking, we don't even assume that our Hamiltonian is from a fermionic system.

I think you mean, "if we have a fermionic system, do we need to assume a JW transformation in order for these quantities to truly be MP denominators"? And yes, that's correct.

The code that introduces the assumption is the [`build_orb_energies` function](https://github.com/evangelistalab/qforte/blob/886de7e3c092b2f11decedf31fedff73ebfb9082/src/qforte/abc/ansatz.py#L54-L56), which is only correct in some JW transformation. That was true before this PR.

Let's assume `build_orb_energies` did not introduce this assumption. Does this code deepen it? Unfortunately, yes. It explicitly assumed that every bit corresponds to some orbital.

I thought about the question of how to make something truly general, and I didn't find a satisfying answer.

src/qforte/ucc/spqe.py
```diff
@@ -16,8 +16,8 @@
 
 class SPQE(UCCPQE):
     """This class implements the selected projective quantum eigensolver (SPQE) for
-    disentagled UCC like ansatz.
-    In SPQE, a batch of imporant particle-hole operators
+    disentangled UCC like ansatz.
+    In SPQE, a batch of important particle-hole operators
     :math:`\{ e^{t_\mu (\hat{\\tau}_\mu - \hat{\\tau}_\mu^\dagger )} \}` are
     added at each macro-iteration :math:`n` to the SPQE unitary :math:`\hat{U}(\mathbf{t})`,
     wile all current parameters are optemized using the quasi-Newton PQE update
```
imagoulas, line 23: "wile" => "while"
"optemized" => "optimized"

src/qforte/ucc/uccnpqe.py
```diff
@@ -18,6 +18,7 @@
 
 import numpy as np
 from scipy.linalg import lstsq
+from collections import Counter
```
imagoulas, line 21: Is Counter used in this module?

src/qforte/ucc/uccnvqe.py
```diff
@@ -259,10 +256,17 @@ def get_num_ham_measurements(self):
...
         #     return 0
         return 0
 
+    def fill_excited_dets(self):
+        reference_state = qforte.QubitBasis(self._nqb)
+        for k, occ in enumerate(self._ref):
+            reference_state.set_bit(k, occ)
+
+        self._excited_dets = [operator_to_determinant(qubit_operator, reference_state) for _, qubit_operator in self._qubit_pool]
+
```
imagoulas, line 271: This function is also defined in uccnpqe.py. Maybe place it in utils to avoid the repetition.

src/qforte/utils/transforms.py
```diff
@@ -327,3 +328,39 @@ def pauli_condense(pauli_op):
         condensed_op.append(condensed_word)
 
     return condensed_op
+
+def operator_to_determinant(operator: qforte.QubitOperator, reference: qforte.QubitBasis) -> tuple[int, complex]:
+
+    """
+    Maps an QubitOperator and a QubitBasis to the index of the basis state
```
imagoulas, line 335: "an" => "a"

imagoulas (approved):
All tests passed successfully on my end. Apart from some typos, I left a couple of comments regarding minor redundancies in the code. Feel free to merge once you are ready.

JonathonMisiewicz:
I'll wait to hear back from Francesco about his comment on fermionic encodings before merging.

src/qforte/abc/ansatz.py
JonathonMisiewicz: @fevangelista Am I clear to merge this PR in, even with the implicit JW assumption?

JonathonMisiewicz:
This PR will need to be completely re-designed to deal with #178 and follow-ups.

PR #160, opened 2021-09-29
QPE Bugfix I

JonathonMisiewicz:
## Description
This PR does some cleanup of the QPE code and also fixes _a_ bug.

To understand the bug, you need only the following rough model of Quantum Phase Estimation:

QPE:
(1) takes in a hermitian Hamiltonian `H` with eigenvalues `λ_m` and a state `Ψ`
(2) constructs `exp(i*H)`, which has eigenvalues `exp(i*λ_m)`
(3) measures the `exp(i*λ_m)` with probability distribution given by your input state
(4) uses the most probable `exp(i*λ_m)` to back out its `λ_m`

The problem is that `exp(i*λ_m)` is periodic in `λ_m` with intervals of `2π`, so you can only determine λ_m up to a factor of 2π. The previous implementation arbitrarily chose the branch with `[-2π, 0]`. This passed tests because the toy system of H2 had a small eigenvalue. Systems with larger eigenvalues find themselves on the wrong branch.

As a remedy, there is now a new required keyword: guess_energy. QPE will choose to add the number of `2π` to get the `λ_m` as close as possible to guess_energy. This fixes _a_ buggy test case, but there may be more.

## User Notes
- [x] QPE now requires a `guess_energy` to convert from a phase to an energy. This is true on exact quantum computers, as well.

## Checklist
- [x] Added tests to newly working features
- [x] Ready to go!

### DIFF
--- a/src/qforte/qpea/qpe.py
+++ b/src/qforte/qpea/qpe.py
@@ -15,6 +15,7 @@ from scipy import stats
 
 class QPE(Algorithm):
     def run(self,
+            guess_energy,
             t = 1.0,
             nruns = 20,
             success_prob = 0.5,
@@ -39,6 +40,9 @@ class QPE(Algorithm):
 
         self._n_classical_params = 0
         self._n_pauli_trm_measures = nruns
+
+        self._guess_energy = guess_energy
+        self._guess_periods = round(self._t * guess_energy / (-2 * np.pi))
 
         # Print options banner (should done for all algorithms).
         self.print_options_banner()
@@ -70,11 +74,7 @@ class QPE(Algorithm):
 
         self._phases = []
         for readout in z_readouts:
-            val = 0.0
-            i = 1
-            for z in readout:
-                val += z / (2**i)
-                i += 1
+            val = sum(z / (2**i) for i, z in enumerate(readout, start=1))
             self._phases.append(val)
 
         # find final binary string of phase readouts:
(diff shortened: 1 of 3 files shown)


imagoulas:
Everything looks good, but, based on the review that you posted on Slack yesterday (page 21), there is an alternative way to fix the bug. Currently, you use an initial guess to find the number of 2π cycles that need to be added to the final phase to bring the final QPE energy as closer to a meaningful result as possible.

An alternative approach would be to shift and scale the qubit Hamiltonian itself, making the resulting eigenvalues lie in the desired phase interval 0 < θ < 1. For this approach to work, we need two energies; a lower/upper bound to the lowest/highest eigenvalue of the Hamiltonian. One way that I came up with (not the best, this is just an idea) to compute these energies is as follows. Assuming that a mean field solution is already available, we let all electrons occupy the lowest energy orbital and compute the energy. This should give as a lower bound to the ground state. Then, we let all electrons occupy the highest energy orbital and compute the energy. Of course, this strategy will fail in severe cases of strong correlation where all orbitals become almost degenerate.

Which one of the two approaches do you feel is more robust and minimizes the danger of getting an erroneous QPE energy?

imagoulas:
In that case, I would suggest either putting a default value for the initial guess, such as the hf_energy in the case of molecules, or raise an error if the user does not provide an initial guess.

JonathonMisiewicz:
> raise an error if the user does not provide an initial guess.

I'm confused. `guess_energy` is not given a default value, so Python does this automatically.

imagoulas:
I was thinking of putting a more meaningful error message than the default one. Something like "The QPE implementation in QForte requires an initial guess to be provided.\n A sensible initial guess would be the HF energy.", or something like that. I am not sure whether other implementations of QPE require initial guesses, so it might be a good idea to guide the user.

JonathonMisiewicz:
A custom error message isn't the standard solution. The standard solution is to use the default error message and have the method properly documented so that a confused user can see what the expected argument is. We need proper documentation anyways, so having the custom error message would be redundant.

_That said_, the documentation (or lack thereof) is a major problem throughout all of QForte, and one I can't commit to fixing at this time between the many other things on my to-do list. I can write a cursory docstring up, but we'll need to revisit it later anyways.

imagoulas:
I totally agree with you. The lack of a proper and thorough documentation is the root of most, if not all, possible issues that a new user might experience. As I said before, the code looks good to me. Feel free to merge whenever you are ready.

imagoulas (approved):
Approved!

PR #162, opened 2021-10-05
Closes 151

### LINKED ISSUE #151: Frozen QPE
Attempts to use frozen core with QPE apparently lead to anomalous energies. @imagoulas would have further details.

We should identify the problem and fix it so that frozen orbitals can be used with QPE.

JonathonMisiewicz:
## Description
Closes #151.

## User Notes
- [x] Frozen orbitals with QPE are now supported.

## Checklist
- [x] Added tests for newly working features
- [x] Ready to go!

### DIFF
--- a/src/qforte/abc/algorithm.py
+++ b/src/qforte/abc/algorithm.py
@@ -75,10 +75,6 @@ class Algorithm(ABC):
                  print_summary_file=False,
                  **kwargs):
 
-        if isinstance(self, qf.QPE) and hasattr(system, 'frozen_core'):
-            if system.frozen_core + system.frozen_virtual > 0:
-                raise ValueError("QPE with frozen orbitals is not currently supported.")
-
         self._sys = system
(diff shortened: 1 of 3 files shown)


imagoulas (changes requested):
Currently, there are 3 tests that fail on my end.

In line 50 of utils/trotterization.py there is a `troterized_operator` that needs to be replaced by `trotterized_operator`. This will fix 2 of the 3 failing tests.

The third failing test is `test_freeze_orb_qpe`. I am getting
`assert -28.736737051008387 == -28.747184707540754 ± 1.1e-03`
Maybe you will need to play with the QPE parameters a little bit to make it pass.

JonathonMisiewicz:
I swear I had the `test_freeze_orb_qpe` test working this morning.

Most of those are trivial, but I'm seeing evidence of _another_ bug in the code. I'll deal with it, although I'm not happy that the situation was this bad. Thanks for catching that something was amiss.

PR #174, opened 2023-02-06
Removed get_final_energy function and modified conv_status

imagoulas:
## Description
Simplified the SPQE code. Removed the get_final_energy function. The modified conv_status function checks whether:
1) SPQE has converged
2) The maximum number of macro-iterations has been reached
3) The operator pool has been completely drained
In any case, it prints the current norm of the residual vector. To accomplish this, some reordering of printing statements was needed.

In a future PR, some related attributes, currently used by other parts of the code, will also be removed.

## User Notes
- [ ] Features added

## Checklist
- [ ] Documented source code
- [x ] Ready to go!

### DIFF
--- a/src/qforte/ucc/spqe.py
+++ b/src/qforte/ucc/spqe.py
@@ -79,6 +79,7 @@ class SPQE(UCCPQE):
         self._grad_norms = []
         self._tops = []
         self._tamps = []
+        self._stop_macro = False
         self._converged = False
         self._res_vec_evals = 0
         self._res_m_evals = 0
@@ -114,7 +115,7 @@ class SPQE(UCCPQE):
             f.write(f"#{'Iter(k)':>8}{'E(k)':>14}{'N(params)':>17}{'N(CNOT)':>18}{'N(measure)':>20}\n")
             f.write('#-------------------------------------------------------------------------------\n')
 
-        while not self._converged:
+        while not self._stop_macro:
 
             self.update_ansatz()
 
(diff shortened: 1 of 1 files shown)


JonathonMisiewicz:
I don't agree with (2). If the maximum number of macro-iterations have been reached without converging, I would normally consider that a failure to converge. I can see some cases (especially for private tests) where you would still mark a success when convergence stalls out. Can we add an option to allow _both_ behaviors, with maximum iterations reached being marked a non-convergence, by default?

imagoulas:
I see what you are saying. I was using the "_converged" in  a much more liberal way, namely, do I need more macro-iterations or not.  What if we add a second attribute such as "_stop_macro"? In that case, we can replace "while not self._converged" by "while not self._stop_macro".

JonathonMisiewicz:
Yes, that sounds agreeable to me!

src/qforte/ucc/spqe.py
```diff
@@ -106,17 +107,16 @@ def run(self,
             self._pool_obj.add_term(0.0, self.get_op_from_basis_idx(I))
 
         self.build_orb_energies()
-        spqe_iter = 0
-        hit_maxiter = 0
+
+        self._spqe_iter = 1
```
JonathonMisiewicz, line 111: Why is this now a member variable?

JonathonMisiewicz (commented):
Okay, much better.

src/qforte/ucc/spqe.py
imagoulas: I decided instead of passing it on the few functions that need it, to make it a member variable. If you prefer passing it, I can do that.

JonathonMisiewicz: At present, I don't see any functions that need it. If another PR is going to add them, I can keep it as a member variable for now.

PR #176, opened 2023-02-06
Symmetry adapted spqe pool

imagoulas:
## Description
Added symmetry-adapted (particle number, Sz, spatial) operator pool in SPQE. To do this efficiently, I created two dictionaries and a list. The "excitation_dictionary" maps the integers representing the symmetry-adapted determinants to the indices of the corresponding second-quantized operators in the pool. The "reversed_excitation_dictionary" is the inverse map. The "excitation_indices" list contains the integers representing the symmetry-adapted determinants. These structures also allowed the elimination of two functions ("add_from_basis_idx" and "get_op_from_basis_idx"). The symmetry-adapted pool permitted the simplification of the "get_residual_vector" function.

## User Notes
- [ ] Features added

## Checklist
- [ ] Documented source code
- [ x] Ready to go!

### DIFF
--- a/src/qforte/ucc/spqe.py
+++ b/src/qforte/ucc/spqe.py
@@ -436,13 +436,15 @@ class SPQE(UCCPQE):
                     print('     op index (Imu)     Number of times measured')
                     print('  -----------------------------------------------')
 
-                for Nmu_tup in Nmu_lst[:-1]:
+                for rmu_sq, global_op_idx in Nmu_lst[:-1]:
                     if(self._verbose):
-                        print(f"  {Nmu_tup[1]:10}                  {np.real(Nmu_tup[0]):14}")
-                    if(Nmu_tup[1] not in self._tops):
-                        self._tops.insert(0,self._excitation_dictionary[Nmu_tup[1]])
+                        print(f"  {global_op_idx:10}                  {np.real(rmu_sq):14}")
+                    if(global_op_idx not in self._tops):
+                        pool_idx = self._excitation_dictionary[global_op_idx]
+                        self._tops.insert(0,pool_idx)
                         self._tamps.insert(0,0.0)
-                        self._nbody_counts[len(self._pool_obj[self._excitation_dictionary[Nmu_tup[1]]][1].terms()[0][1]) - 1] += 1
+                        operator_rank = len(self._pool_obj[pool_idx][1].terms()[0][1])
+                        self._nbody_counts[operator_rank - 1] += 1
 
                 self._n_classical_params_lst.append(len(self._tops))
 
@@ -468,15 +470,17 @@ class SPQE(UCCPQE):
                     # Make a running list of operators. When the sum of res_sq exceeds the target, every operator
                     # from here out is getting added to the ansatz..
                     temp_ops = []
-                    for rmu_sq, op_idx in res_sq:
+                    for rmu_sq, global_op_idx in res_sq:
                         res_sq_sum += rmu_sq / (self._dt * self._dt)
                         if res_sq_sum > (self._spqe_thresh * self._spqe_thresh):
+                            pool_idx = self._excitation_dictionary[global_op_idx]
                             if(self._verbose):
-                                print(f"  {self._excitation_dictionary[op_idx]:10}                  {np.real(rmu_sq):14.12f}"
-                                      f"   {self._pool_obj[self._excitation_dictionary[op_idx]][1].str()}" )
-                            if self._excitation_dictionary[op_idx] not in self._tops:
-                                temp_ops.append(self._excitation_dictionary[op_idx])
-                                self._nbody_counts[len(self._pool_obj[self._excitation_dictionary[op_idx]][1].terms()[0][1]) - 1] += 1
+                                print(f"  {pool_idx:10}                  {np.real(rmu_sq):14.12f}"
+                                      f"   {self._pool_obj[pool_idx][1].str()}" )
+                            if pool_idx not in self._tops:
+                                temp_ops.append(pool_idx)
+                                operator_rank = len(self._pool_obj[pool_idx][1].terms()[0][1])
+                                self._nbody_counts[operator_rank - 1] += 1
 
                     for temp_op in temp_ops[::-1]:
                         self._tops.insert(0, temp_op)
@@ -485,13 +489,15 @@ class SPQE(UCCPQE):
                 else:
                     # Add the single operator with greatest rmu_sq not yet in the ansatz
                     res_sq.reverse()
-                    for rmu_sq, op_idx in res_sq[1:]:
-                        print(f"  {self._excitation_dictionary[op_idx]:10}                  {np.real(rmu_sq)/(self._dt * self._dt):14.12f}")
-                        if self._excitation_dictionary[op_idx] not in self._tops:
+                    for rmu_sq, global_op_idx in res_sq:
+                        pool_idx = self._excitation_dictionary[global_op_idx]
+                        print(f"  {pool_idx:10}                  {np.real(rmu_sq)/(self._dt * self._dt):14.12f}")
+                        if pool_idx not in self._tops:
                             print('Adding this operator to ansatz')
-                            self._tops.insert(0, self._excitation_dictionary[op_idx])
+                            self._tops.insert(0, pool_idx)
                             self._tamps.insert(0, 0.0)
-                            self._nbody_counts[len(self._pool_obj[self._excitation_dictionary[op_idx]][1].terms()[0][1]) - 1] += 1
+                            operator_rank = len(self._pool_obj[pool_idx][1].terms()[0][1])
+                            self._nbody_counts[operator_rank - 1] += 1
                             break
 
                 self._n_classical_params_lst.append(len(self._tops))
--- a/src/qforte/ucc/spqe.py
+++ b/src/qforte/ucc/spqe.py
@@ -103,7 +103,9 @@ class SPQE(UCCPQE):
                 self._nbody_counts.append(0)
 
         # create a pool of particle number, Sz, and spatial symmetry adapted second quantized operators
+        # Encode the occupation list into a bitstring
         ref = sum([b << i for i, b in enumerate(self._ref)])
+       # `& mask_alpha` gives the alpha component of a bitstring. `& mask_beta` does likewise.
         mask_alpha = 0x5555555555555555
         mask_beta = mask_alpha << 1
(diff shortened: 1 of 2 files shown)


JonathonMisiewicz:
Please add a test case that the SPQE operator pool is working correctly.

src/qforte/ucc/spqe.py
```diff
@@ -101,9 +102,41 @@ def run(self,
             if occupation:
                 self._nbody_counts.append(0)
 
+        # create a pool of particle number, Sz, and spatial symmetry adapted second quantized operators
+        ref = sum([b << i for i, b in enumerate(self._ref)])
```
JonathonMisiewicz, line 106:
```suggestion
        # Encode the occupation list into a bitstring
        ref = sum([b << i for i, b in enumerate(self._ref)])
```

src/qforte/ucc/spqe.py
```diff
@@ -101,9 +102,41 @@ def run(self,
...
+        mask_alpha = 0x5555555555555555
```
JonathonMisiewicz, line 107:
```suggestion
       # `& mask_alpha` gives the alpha component of a bitstring. `& mask_beta` does likewise.
        mask_alpha = 0x5555555555555555
```

src/qforte/ucc/spqe.py
```diff
@@ -101,9 +102,41 @@ def run(self,
...
+        idx = 0
+        # determinant id : excitation operator index
+        self._excitation_dictionary = {}
+        # list that holds the ids of determinants corresponding to operators in the pool
+        self._excitation_indices = []
         self._pool_obj = qf.SQOpPool()
-        for I in range(2 ** self._nqb):
-            self._pool_obj.add_term(0.0, self.get_op_from_basis_idx(I))
+        for I in range(1 << self._nqb):
+            alphas = [int(j) for j in bin(I & mask_alpha)[2:]]
```
JonathonMisiewicz, line 119: Why is the `[2:]` necessary?

src/qforte/ucc/spqe.py
```diff
@@ -101,9 +102,41 @@ def run(self,
...
+            betas = [int(j) for j in bin(I & mask_beta)[2:]]
+            if sum(alphas) == nalpha and sum(betas) == nbeta:
```
JonathonMisiewicz, line 121:
```suggestion
            # Confirm Sz symmetry
            if sum(alphas) == nalpha and sum(betas) == nbeta:
```

src/qforte/ucc/spqe.py
```diff
@@ -101,9 +102,41 @@ def run(self,
...
+                if sq_op_find_symmetry(self._sys.orb_irreps_to_int,
```
JonathonMisiewicz, line 122:
```suggestion
                # Confirm point group symmetry
                if sq_op_find_symmetry(self._sys.orb_irreps_to_int,
```

src/qforte/ucc/spqe.py
```diff
@@ -101,9 +102,41 @@ def run(self,
...
+                                       [len(alphas) - i - 1 for i, x in enumerate(alphas) if x],
+                                       [len(betas) -i - 1 for i, x in enumerate(betas) if x]) == self._irrep:
+                    excit = bin(ref ^ I).replace("0b", "")
+                    if excit != "0":
```
JonathonMisiewicz, line 126:
```suggestion
                    # Confirm excitation number is non-zero
                    if excit != "0":
```

src/qforte/ucc/spqe.py
```diff
@@ -101,9 +102,41 @@ def run(self,
...
-        for I in range(2 ** self._nqb):
-            self._pool_obj.add_term(0.0, self.get_op_from_basis_idx(I))
+        for I in range(1 << self._nqb):
+            alphas = [int(j) for j in bin(I & mask_alpha)[2:]]
+            betas = [int(j) for j in bin(I & mask_beta)[2:]]
+            if sum(alphas) == nalpha and sum(betas) == nbeta:
+                if sq_op_find_symmetry(self._sys.orb_irreps_to_int,
+                                       [len(alphas) - i - 1 for i, x in enumerate(alphas) if x],
+                                       [len(betas) -i - 1 for i, x in enumerate(betas) if x]) == self._irrep:
+                    excit = bin(ref ^ I).replace("0b", "")
```
JonathonMisiewicz, line 125:
```suggestion
                   # Create the bitstring of created/annihilated orbitals
                    excit = bin(ref ^ I).replace("0b", "")
```

src/qforte/ucc/spqe.py
```diff
@@ -309,59 +342,17 @@ def get_residual_vector(self, trial_amps):
...
-                I = basis_I.add()
 
-                ## check for correct dets
-                det_I = integer_to_ref(I, self._nqb)
-                nel_I = sum(det_I)
-                cor_spin_I = correct_spin(det_I, 0)
+            qc_temp = qforte.Computer(self._nqb)
+            qc_temp.apply_circuit(self._Uprep)
+            qc_temp.apply_operator(sq_op.jw_transform())
+            sign_adjust = qc_temp.get_coeff_vec()[self._reversed_excitation_dictionary[m]]
```
JonathonMisiewicz, line 349: Why, exactly is `sign_adjust` needed? That isn't clear to me from the original code.

src/qforte/ucc/spqe.py
```diff
@@ -439,16 +430,16 @@ def update_ansatz(self):
                     if(self._verbose):
                         print(f"  {Nmu_tup[1]:10}                  {np.real(Nmu_tup[0]):14}")
                     if(Nmu_tup[1] not in self._tops):
-                        self._tops.insert(0,Nmu_tup[1])
+                        self._tops.insert(0,self._excitation_dictionary[Nmu_tup[1]])
                         self._tamps.insert(0,0.0)
-                        self.add_from_basis_idx(Nmu_tup[1])
+                        self._nbody_counts[len(self._pool_obj[self._excitation_dictionary[Nmu_tup[1]]][1].terms()[0][1]) - 1] += 1
```
JonathonMisiewicz, line 435: What is `_nbody_counts`? It's not documented.

src/qforte/ucc/spqe.py
```diff
@@ -439,16 +430,16 @@ def update_ansatz(self):
...
 
                 self._n_classical_params_lst.append(len(self._tops))
 
         else: # when M_omega == 'inf', proceed with standard SPQE
```
JonathonMisiewicz, line 439: What's `M_omega`? That's nto documented, either.

src/qforte/ucc/spqe.py
imagoulas: Let's assume that we have M spin-orbitals. Using `& mask_alpha` on a given M-bit string will return an M-bit string with 0 in the "beta" positions and the alpha positions unchanged. I want the "alphas" list to have length equal to the number of spatials (M/2), hence the `[2:]` part.

src/qforte/ucc/spqe.py
imagoulas: I had the same question as you. In fact, when I first saw that I tried to remove it. The numerical results did not make sense. At some point I believe I figured out why it was needed, but now this knowledge has evaporated.

src/qforte/ucc/spqe.py
imagoulas: _nbody_counts is a list that stores the numbers of singles, doubles, .... in the final ansatz. It is being printed near the end of an SPQE run. I will add a comment with the explanation,

src/qforte/ucc/spqe.py
imagoulas: M_omega is the total number of measurements for determining the residuals (see the M symbol in Eq. (16) of the original SPQE paper). Essentially, setting it to "inf" means perfect measurement of all residuals. I never played with anything else than "inf". I will add a comment with the explanation.

src/qforte/ucc/spqe.py
JonathonMisiewicz: I just tested this. I think you mean that you want to strip the "0b" prefix?

src/qforte/ucc/spqe.py
JonathonMisiewicz: I'll try to figure it out myself.

JonathonMisiewicz: Got it.
coeffs_I is U^HU|Phi>. We want to measure <Phi| K U^HU|Phi>. We know which vector element of coeffs_I to probe, but the corresponding basis element might not have the same parity as <Phi |K. We correct for that by checking the parity of the corresponding basis element in K|Phi>.
If that makes sense, this _really_ should be a comment.

src/qforte/ucc/spqe.py
imagoulas: You are correct. I thought I was looping every two elements, but I was wrong.

JonathonMisiewicz: That'd be [::2], so not too far off.

src/qforte/ucc/spqe.py
imagoulas: Yep. Fermionic antisymmetry. I will add a comment.

src/qforte/ucc/spqe.py
```diff
@@ -439,16 +440,16 @@ def update_ansatz(self):
                     if(self._verbose):
                         print(f"  {Nmu_tup[1]:10}                  {np.real(Nmu_tup[0]):14}")
                     if(Nmu_tup[1] not in self._tops):
-                        self._tops.insert(0,Nmu_tup[1])
+                        self._tops.insert(0,self._excitation_dictionary[Nmu_tup[1]])
```
JonathonMisiewicz, line 443: This part of the code was confusing to begin with, and the sheer length of `len(self._pool_obj[self._excitation_dictionary[Nmu_tup[1]]][1].terms()[0][1]) - 1` makes it worse. Might I suggest writing `_, global_op_idx = Nmu_tup` so that you can replace `Nmu_tup[1]`? You can also define `pool_idx = self._excitation_dictionary[global_op_idx]`. I'd also encourage defining `operator_rank = len(self._pool_obj[pool_idx][1].terms()[0][1]) - 1` so that the incrementing of `_nbody_counts` is more transparent. These auxiliary variables should clean up most of the below code, as well.

Good variable names are self-documenting!

JonathonMisiewicz (commented):
I have one other request for code readability.

JonathonMisiewicz (approved):

PR #178, opened 2023-02-07
Added qubit excitations

imagoulas:
## Description
The jw_transform attribute has a new, boolean, optional argument allowing for qubit excitations (essentially neglecting the strings of Z gates). the test_jw_transform.py file has been updated with a new test confirming the new functionality.

## User Notes
- [ x] Features added

## Checklist
- [ x] Added/updated tests of new features
- [ ] Documented source code
- [ x] Ready to go!

### DIFF
--- a/src/qforte/abc/algorithm.py
+++ b/src/qforte/abc/algorithm.py
@@ -224,6 +224,9 @@ class AnsatzAlgorithm(Algorithm):
     _tamps : list
         A list of amplitudes (to be optimized) representing selected
         operators in the pool.
+
+    _qubit_excitations: bool
+        Controls the use of qubit/fermionic excitations.
     """
 
     # TODO (opt major): write a C function that prepares this super efficiently
@@ -294,13 +297,15 @@ class AnsatzAlgorithm(Algorithm):
 
         return val
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, *args, qubit_excitations=False,
+            **kwargs):
         super().__init__(*args, **kwargs)
         self._curr_energy = 0
         self._Nm = []
         self._tamps = []
         self._tops = []
         self._pool_obj = qf.SQOpPool()
+        self._qubit_excitations = qubit_excitations
 
         kwargs.setdefault('irrep', None)
         if hasattr(self._sys, 'point_group'):
--- a/src/qforte/abc/uccvqeabc.py
+++ b/src/qforte/abc/uccvqeabc.py
@@ -145,7 +145,7 @@ class UCCVQE(VQE, UCC):
         mu = M-1
 
         # find <sing_N | K_N | psi_N>
-        Kmu_prev = self._pool_obj[self._tops[mu]][1].jw_transform()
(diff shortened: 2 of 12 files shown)


JonathonMisiewicz:
I'll have more detailed thoughts tomorrow, but storing qubit operators in an `SqOpPool` strikes me as a bad idea.

JonathonMisiewicz:
@imagoulas Please resolve merge conflicts.

imagoulas:
> 

In general, quite a lot. These are the PRs that come to my mind right now:
CNOT-efficient fermionic/qubit excitations
Fixing DIIS and modularize Jacobi and DIIS
Add residual minimization for SPQE
Add spin operator functionality
Add symmetry analysis function
Add non-iterative corrections based on moments
Add maximum excitation rank option for SPQE pool

JonathonMisiewicz:
Qubit operators are not second quantized operators, but you are modifying `SQOpPool` to refer to a set of _site_ creators and annihilators without regard for whether they're SQ or Qubit creators/annihilators. `SQOperator::JW_transform` is not even acting on a set of _second quantized_ creators and annihilators but on a set of _qubit_ creators and annihilators. That's quite a problem for the `SQOperator` class!

I'm not comfortable merging this to master. We can either fix this now and aim to merge for master or admit that we'll need to deal with the kludge later and merge it into a separate branch. I'm inclined to the _latter_. Any solution I propose may cause problems for you later.

JonathonMisiewicz:
Okay, let's take a step back.

The mapping from second quantized operators to qubit operators given by "second quantized create/annihilate p to qubit create/annihilate p" does not arise from _any_ fermionic encoding for more than one electron. Given two distinct sites `r` and `s`, which double as both orbitals and qubits, we know `a†r a†s = - a†s a†r`, but `q†r q†s = q†s q†r`. Your JW transform is therefore not well-defined on SQ operators, only on strings of non-commuting letters.

Essentially, this change forces `SQOperator` to no longer represent a bona-fide SQ operator, but a string of letters where no commutation relations can be assumed. These letters can either be interpreted as SQ creation/annihilation or qubit creation/annihilation (your code adds this second interpretation).

At the very least, this PR means we need to rename the `SQOpPool` and `SqOperator` classes, but I'd need to check both those classes carefully to see if we're creating other inconsistencies, now that the class can't automatically assume fermionic commutation relations. That's going to be a long and complicated process, and it may be best to merge to a separate branch before we decide whether it's enough to rename those classes, or whether more invasive measures are needed.

imagoulas:
I still do not see the issue. The second quantized operators are not modified. The anticommutation relations still hold as they should. The way I see it is that what changes is the kind of mapping between second quantized operators and their "image" in the qubit space. The standard JW map preserves the structure (homomorphism) and the resulting qubit operators anticommute. The modified JW map does not preserve the structure (morphism?) since now the resulting qubit operators obey a different algebra. If anything, I would say that QubitOperator and QubitOpPool may need to be renamed into something more abstract, as they may be confused with qubit excitation operators.

We can speed up the process via a Zoom.

JonathonMisiewicz (approved):
This works as a temporary solution. Removing the fermionic SQ assumption is going to be important for future development.

PR #184, opened 2023-02-13
Moment energy corrections

imagoulas:
## Description
Added non-iterative energy corrections based on the method of moments of coupled-cluster theory. So far, two types of corrections have been considered, the first using Moller-Plesset and the second Epstein-Nesbet denominators. These corrections are currently interfaced with all ansatz-dependent algorithms in QForte, namely, UCCNPQE, SPQE, UCCNVQE, and ADAPTVQE.

A test for the new functionality has been added.

## User Notes
- [x ] Features added

## Checklist
- [x ] Added/updated tests of new features
- [x ] Documented source code
- [x ] Ready to go!

### DIFF
--- a/src/qforte/utils/moment_energy_corrections.py
+++ b/src/qforte/utils/moment_energy_corrections.py
@@ -17,7 +17,8 @@ def construct_moment_space(self):
     self._mmcc_excitation_indices = []
     self._mmcc_pool = qf.SQOpPool()
 
-    self.build_orb_energies()
+    if not hasattr(self, '_orb_e'):
+        self.build_orb_energies()
 
     # create a pool of particle number, Sz, and spatial symmetry adapted second quantized operators
     # Encode the occupation list into a bitstring
@@ -30,7 +31,7 @@ def construct_moment_space(self):
 
     if self._max_moment_rank is None:
         self._max_moment_rank = nalpha + nbeta
-    elif not isinstance(self._max_moment_rank, int) or max_momnet_rank <= 0:
+    elif not isinstance(self._max_moment_rank, int) or self._max_moment_rank <= 0:
         raise TypeError("The maximum moment rank must be a positive integer!")
     elif self._max_moment_rank > nalpha + nbeta:
         # WARNING: This step could be potetnially removed, if QForte gets methods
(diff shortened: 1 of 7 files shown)


JonathonMisiewicz (approved):
Like I said privately, I just can't give this a more thorough review without more theoretical knowledge. A comment saying this is experimental, publication forthcoming, would be for the best.

src/qforte/utils/moment_energy_corrections.py
```diff
@@ -0,0 +1,99 @@
...
+        self._epstein_nesbet.append(qc.direct_op_exp_val(self._qb_ham))
+
+def compute_moment_energies(self):
+
+    # The moments, i.e., residuals, are estimated by measuring the "residual" state a la SPQE
+
+    if hasattr(self, '_dt'):
+        dt = self._dt
+    else:
+        # not sure about this
```
JonathonMisiewicz, line 79: What is "this"? The wisdom of this particular choice of `self._dt`? Having this attribute be `_dt` in the first place?

src/qforte/utils/moment_energy_corrections.py
```diff
@@ -0,0 +1,99 @@
...
+    nbeta = sum(self._ref[1::2])
+
+    if not isinstance(self._max_moment_rank, int) or self._max_moment_rank < 0:
+        raise TypeError("The maximum moment rank must be a non-negative integer!")
+    elif self._max_moment_rank > nalpha + nbeta:
+        # WARNING: This step could be potetnially removed, if QForte allows for methods
+        #          that violate particle number symmetry
+        self._max_moment_rank = nalpha + nbeta
+
+    for I in range(1 << self._nqb):
```
JonathonMisiewicz, line 39: Can we factor out this symmetry selection code? I remember similar code elsewhere in this PR push.

src/qforte/utils/moment_energy_corrections.py
imagoulas: In the case of SPQE, there exists an attribute called "_dt" and the user has the ability to change its value from the default of 0.001. The other algorithms, namely, UCCNPQE, UCCNVQE, and ADAPTVQE, do not have this attribute and creating it seems confusing and unnecessary. For such cases, the "dt" variable is hard-coded to 0.001 (the default for SPQE). The "not sure about this" refers to the hard-coding part. I am open to suggestions :-)

src/qforte/utils/moment_energy_corrections.py
imagoulas: There are some minor differences between the two codes, but we should be able to factor it out.

JonathonMisiewicz: Agreed privately, that this code isn't stable enough to factor out yet.

src/qforte/utils/moment_energy_corrections.py
JonathonMisiewicz: Oh. I'd create a new variable rather than re-use `_dt`. Maybe call it `_moment_dt`. The two do (slightly) different things.

JonathonMisiewicz: Leave it flexible.

JonathonMisiewicz (approved):

PR #187, opened 2023-02-14
Symmetry analysis

imagoulas:
## Description
Added a symmetry analysis function. For a given quantum state, it computes the expectation values and variances of the particle number N, projection of total spin on the z axis Sz, and total spin squared S^2 operators. It also computes the weight contribution to the wavefunction of the various N, Sz, and spatial symmetry sectors.

A test of this functionality will be included after the implementation of the approximate FEB/QEB circuits, which break N and Sz symmetries.

In an initial commit, I accidentally included modifications to the code pertaining to the energy shift of Moller-Plesset denominators in Jacobi iterations. These additions have been removed.

## User Notes
- [x ] Features added

## Checklist
- [x ] Documented source code
- [x ] Ready to go!

### DIFF
--- a/src/qforte/utils/symmetry_analysis.py
+++ b/src/qforte/utils/symmetry_analysis.py
@@ -1,7 +1,7 @@
 import qforte as qf
 import numpy as np
 
-def symmetry_check(n_qubits, Uprep, pool, tops, tamps, compact, qubit, trotter_number, irreps, orb_irreps_to_int):
+def symmetry_check(n_qubits, qc, irreps, orb_irreps_to_int):
     """
     This function performs a symmetry analysis of the provided quantum state.
     In particular, it computes the expectation values and variances of the
@@ -13,28 +13,10 @@ def symmetry_check(n_qubits, Uprep, pool, tops, tamps, compact, qubit, trotter_n
     =========
 
     n_qubits: int
-              The number of qubits of the system.
+        The number of qubits of the system.
 
-    Uprep: Circuit
-           The circuit that transforms the |0..0> state to the reference one.
-
-    pool: SQOpPool
-        Pool of second-quantized operators used in constructing the ansatz.
-
-    tops: list of int
-        List of indices denoting the operators from the pool that have been selected for the ansatz.
-
-    tamps: list of float
-        Amplitudes corresponding to the second-quantized operators in tops.
-
-    compact: boolean
-        Controls the use of standard vs compact excitation circuits.
-
-    qubit: boolean
-        Controls the use of fermionic vs qubit anti-Hermitian excitations.
-
-    trotter_number: int
-        The number of Trotter steps used to approximate the ansatz. Currently implemented only for compact = False.
+    qc: Computer
+        The quantum computer prepared in the derised state.
 
     irreps: list of str
         List that holds the irreps of the group in the Cotton ordering.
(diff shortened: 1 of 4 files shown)


src/qforte/utils/symmetry_analysis.py
```diff
@@ -0,0 +1,170 @@
...
+        List that holds the irreps of the group in the Cotton ordering.
+
+    orb_irreps_to_int: list of int
+        Integer n indexes the irrep of spatial orbital n.
+    """
+
+    # Generate the quantum state defined by the Uprep, tops, tamps, compact, qubit and trotter_number
+    qc = qf.Computer(n_qubits)
+    qc.apply_circuit(Uprep)
+    if compact:
```
JonathonMisiewicz, line 49: I _really_ don't like that this code needs to know the details of how to construct the computer. Can we have the code take in the Computer as an argument, instead?

imagoulas: I do agree with you that the code looks ugly. If we pass the Computer as an argument, whenever we want to use this function we will need to do the following:
```
U = self.ansatz_circuit()
comp = qf.Computer(self._nqb)
comp.apply_circuit(self._Uprep)
comp.apply_circuit(U)
```
Indeed, much more concise. Should I proceed with the simplifications?

JonathonMisiewicz: You can make it even shorter.
```
comp = qf.Computer(self._nqb)
comp.apply_circuit(self.build_Uvqc())
```

Yes, simplify ahead.

src/qforte/utils/symmetry_analysis.py
```diff
@@ -0,0 +1,131 @@
...
+
+    print('\n\n*******************')
+    print('*Symmetry Analysis*')
+    print('*******************')
+
+    print('\nParticle number symmetry analysis')
+    print('=================================')
+
+    for i in range(n_qubits + 1):
+        print('N =', str(i).rjust(2, ' '), 'weight:', f'{weight_particle_number[i]:.10f}')
```
JonathonMisiewicz, line 109: What are these "weights", and why do we print them on a per-qubit basis?

imagoulas: The weights quantify the contribution of each symmetry sector to the final wavefunction. For example, let us assume that the wavefunction of a 6-electron (N = 6) system looks like this:

|Ψ> = c_1 |Φ_1> + c_2 |Φ_2> + c_3 |Φ_3> + c_4 |Φ_4> + c_5 |Φ_5> + c_6 |Φ_6> + d_1 |D_1> + d_2 |D_2>,

where  the last two determinants are contaminants containing N = 5 electrons. We thus have the following weights:

weight(N = 5) =  |d_1|^2 + |d_2|^2
weight(N = 6) =  |c_1|^2 + |c_2|^2 + |c_3|^2 + |c_4|^2 + |c_5|^2 + |c_6|^2

Normalization requires:

weight(N = 5) + weight(N = 6) = 1

They are not printed per qubit, but per total particle number, N in [0, n_qubits].

The code works in a similar manner for the Sz and spatial symmetries. Since we are using a basis of Slater determinants, it is not straightforward to do a similar analysis for S^2.

src/qforte/utils/symmetry_analysis.py
```diff
@@ -0,0 +1,131 @@
...
+        weight = qc.get_coeff_vec()[det]
+        weight = weight * np.conjugate(weight)
+        particle_number = len(occ_alpha + occ_beta)
+        Sz_value = (len(occ_alpha) - len(occ_beta)) * 0.5
+        irrep = qf.sq_op_find_symmetry(orb_irreps_to_int, occ_alpha + occ_beta, [])
+        weight_particle_number[particle_number] += weight
+        weight_Sz[Sz_values_dictionary[Sz_value]] += weight
+        weight_irreps[irrep] += weight
+    for i in range(n_qubits + 1):
+        weight_particle_number[i] = np.real(weight_particle_number[i])
```
JonathonMisiewicz, line 96: l. 88 should guarantee these are real, so why is this necessary? At _most_, just call `np.real` on l. 88.

src/qforte/utils/symmetry_analysis.py
```diff
@@ -0,0 +1,131 @@
...
+    S_sqrd_sqrd.add_op(S_sqrd)
+    S_sqrd_sqrd.operator_product(S_sqrd, True, True)
+    S_sqrd_exp_val = qc.direct_op_exp_val(S_sqrd)
+    S_sqrd_var = qc.direct_op_exp_val(S_sqrd_sqrd) - S_sqrd_exp_val**2
+    if np.imag(S_sqrd_exp_val) !=0 or np.imag(S_sqrd_var) !=0:
+        raise ValueError('The expectation value and variance of S^2 should be real!')
+    S_sqrd_exp_val = np.real(S_sqrd_exp_val)
+    S_sqrd_var = np.real(S_sqrd_var)
+
+    # Compute the contributions of various symmetry spaces to the wavefunction
```
JonathonMisiewicz, line 70:
```suggestion
    # Compute the weight of various symmetry subspaces in the wavefunction
```

src/qforte/utils/symmetry_analysis.py
```diff
@@ -0,0 +1,131 @@
...
+    weight_particle_number = [0] * (n_qubits + 1)
```
JonathonMisiewicz, line 71: I'd rename this variable to `weight_per_particle_number`, and so forth.

JonathonMisiewicz (approved):

JonathonMisiewicz:
Approval here means, "fine to merge after addressing comments."

src/qforte/utils/symmetry_analysis.py
imagoulas: This is done purely for printing purposes: I want to avoid printing 0.0000000000j all the time. I will follow your suggestion and call `np.real` on line 88.

PR #191, opened 2023-02-16
Misc. SPQE Cleanup

JonathonMisiewicz:
## Description
Some miscellaneous SPQE cleanup in preparation for pool generalization: variable renaming, changing a list to a set, collecting code...

There will of course be more to go.

## Checklist
- [x] Ready to go!

### DIFF
--- a/src/qforte/ucc/spqe.py
+++ b/src/qforte/ucc/spqe.py
@@ -154,8 +154,7 @@ class SPQE(UCCPQE):
         idx = 0
         # determinant id : excitation operator index
         self._excitation_dictionary = {}
-        # list that holds the ids of determinants corresponding to operators in the pool
-        self._excitation_indices = []
+        self._indices_of_zeroable_residuals = set()
         self._pool_obj = qf.SQOpPool()
         for I in range(1 << self._nqb):
             alphas = [int(j) for j in bin(I & mask_alpha)[2:]]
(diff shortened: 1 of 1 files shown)


src/qforte/ucc/spqe.py
```diff
@@ -180,7 +179,7 @@ def run(self,
                             sq_op.simplify()
                             self._pool_obj.add_term(0.0, sq_op)
                             self._excitation_dictionary[I] = idx
-                            self._excitation_indices.append(I)
+                            self._indices_of_zeroable_residuals.add(I)
```
imagoulas, line 182: Minor suggestion. Maybe use "_indices_of_residuals_to_be_zeroed". The reasoning is that theyr could be other residuals that can be zeroed but are not included in the pool (especially now that we have a maximum excitation rank criterion).

imagoulas (approved):
I just left a minor suggestion, but I leave it up to you. I am fine either way. Feel free to merge when you are ready.

src/qforte/ucc/spqe.py
JonathonMisiewicz: These aren't indices of residuals that can be zeroed. They are indices of residuals _that the algorithm could possibly enforce to be zero with the current pool, but not necessarily with the current ansatz_.

imagoulas: I understand and I wrote the above suggestion knowing that. If we do not mind being more verbose, we can say "_indices_of_zeroable_residuals_in_pool". I believe that this would clear any possible confusion with zeroable residuals in general.

PR #194, opened 2023-02-16
Rename SPQE variables

JonathonMisiewicz:
## Description
More changing of names of internal variables, for clarity.

## Checklist
- [x] Ready to go!

### DIFF
--- a/src/qforte/ucc/spqe.py
+++ b/src/qforte/ucc/spqe.py
@@ -23,7 +23,7 @@ class SPQE(UCCPQE):
     In SPQE, a batch of imporant particle-hole operators
     :math:`\{ e^{t_\mu (\hat{\\tau}_\mu - \hat{\\tau}_\mu^\dagger )} \}` are
     added at each macro-iteration :math:`n` to the SPQE unitary :math:`\hat{U}(\mathbf{t})`,
-    wile all current parameters are optemized using the quasi-Newton PQE update
+    wile all current parameters are optimized using the quasi-Newton PQE update
     with micro-iterations :math:`k`.
 
     In our selection approach we consider a (normalized) quantum state of the form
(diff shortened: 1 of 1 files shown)


src/qforte/ucc/spqe.py
```diff
@@ -152,8 +155,8 @@ def run(self,
         self._pool_type = max_excit_rank
 
         idx = 0
-        # determinant id : excitation operator index
-        self._excitation_dictionary = {}
+        # Given a coefficient index, what is the index of the "corresponding" pool element? Used to compute the operator to add to the ansatz in macroiterations.
+        self._coeff_idx_to_pool_idx = {}
```
imagoulas, line 159: I understand why you used `coeff_idx`, but perhaps replacing it by `det_idx` or `det_int` might be clearer, especially since in this part of the code the `coeff ` variable has not been defined yet.

The way I think about it is as follows.
Let's say that we are dealing with H6/STO-6G. The HF reference determinant |111111000000> corresponds to the integer 63 (111111). Let's assume that the first operator to be appended to the ansatz is a HOMO -> LUMO double excitation. This operator is associated with the determinant |111100110000>, which corresponds to the integer 207. Thus, given the integer 207 the dictionary will return 0, namely, the first element of the pool.

What do you think?

src/qforte/ucc/spqe.py
```diff
@@ -178,12 +181,13 @@ def run(self,
...
-                            self._excitation_dictionary[I] = idx
+                            self._coeff_idx_to_pool_idx[I] = idx
                             self._indices_of_zeroable_residuals_for_pool.add(I)
                             idx += 1
 
-        # excitation operator index : determinant id
-        self._reversed_excitation_dictionary = {value: key for key, value in self._excitation_dictionary.items()}
+        # Given a pool index, what is the coefficient of the "corresponding" coefficient vector element? Used to extract significant residuals in microiterations.
+        # WARNING! This variable will likely need to be replaced to support generalized pools.
+        self._pool_idx_to_coeff_idx = {value: key for key, value in self._coeff_idx_to_pool_idx.items()}
```
imagoulas, line 190: I am not sure why the variable would need to be replaced. Even for generalized excitations, one should still have a one-to-one correspondence between excitation operators and projections. The trouble would be assigning the generalized excitations to particular determinants, since now there is some freedom. A possible trouble with GSD and PQE, in particular when using minimum basis sets, is that the operator pool is larger than FCI. That would make PQE UCCGSD impossible. unless you select a subset of generalized excitations or come up with a different optimization procedure.

imagoulas (approved):
Again,  a minor suggestion regarding the variable renaming. My second comment does not require any modifications and we can talk about it privately. Feel free to merge.

PR #204, opened 2023-04-06
Enabled the drawing of quantum circuits using latex

imagoulas:
## Description
A new attribute has been added to the Circuit class that allows the drawing of Circuit objects using latex. The quantikz latex package is required for compiling. If the circuit is too large, compiling with luatex is better. Currently, all single-qubit gates are supported and the cZ, CNOT, and aCNOT two-qubit gates.

Not sure if a test of this functionality is needed.

## User Notes
- [x ] Features added

## Checklist
- [x ] Documented source code
- [x ] Ready to go!

### DIFF
--- a/src/qforte/utils/circuit_to_latex.py
+++ b/src/qforte/utils/circuit_to_latex.py
@@ -123,18 +123,23 @@ def circuit_to_latex(circ, filename = 'circuit', max_circuit_depth_per_tikz = 20
             gate_count[gate.target()] += 1
         else:
             # currently QForte supports up to two-qubit gates
-            count_target = gate_count[gate.target()]
-            count_control = gate_count[gate.control()]
-            # if the target and control wires contain a different
-            # number of gates, we need to add "\qw" to equalize them
-            if count_target < count_control:
-                for i in range(count_control - count_target):
-                    wires[gate.target()] += " & \qw"
-                    gate_count[gate.target()] += 1
-            elif count_target > count_control:
-                for i in range(count_target - count_control):
-                    wires[gate.control()] += " & \qw"
-                    gate_count[gate.control()] += 1
+            # To avoid overlapping gates, the gate count of
+            # the target and control qubits as well as all
+            # qubits inbetween are shifted accordingly.
+            if gate.target() > gate.control():
+                max_qubit_id = gate.target()
+                min_qubit_id = gate.control()
+            else:
+                max_qubit_id = gate.control()
+                min_qubit_id = gate.target()
+            max_depth = 0
+            for wire in range(min_qubit_id, max_qubit_id + 1):
+                if max_depth < gate_count[wire]:
+                    max_depth = gate_count[wire]
+            for wire in range(min_qubit_id, max_qubit_id + 1):
+                for i in range(max_depth - gate_count[wire]):
+                    wires[wire] += " & \qw"
+                    gate_count[wire] += 1
             diff = int(gate.control() - gate.target())
             if gate.gate_id() not in ["cZ", "aCNOT", "acX", "CNOT", "cX"]:
                 raise ValueError("The only two-qubit gates that are currently supported are: CNOT, aCNOT, and cZ!")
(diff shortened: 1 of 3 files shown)


src/qforte/utils/circuit_to_latex.py
```diff
@@ -0,0 +1,145 @@
...
+
+    tikz_end = r"""        \end{quantikz}
+    };
+\end{tikzpicture}
+"""
+
+    epilogue = r"""
+\end{document}"""
+
+    texfile = open("circuit.tex", "w")
```
JonathonMisiewicz, line 50: Should the filename be an optional parameter?

src/qforte/utils/circuit_to_latex.py
```diff
@@ -0,0 +1,145 @@
...
+    texfile.write(preamble)
+    texfile.write(tikz_start)
+
+    # The maximum gate count for a given tikz drawing.
+    # This makes compilation easier
+    max_gate_count_per_tikz = 20
```
JonathonMisiewicz, line 56: Should this be an optional parameter?

src/qforte/utils/circuit_to_latex.py
```diff
@@ -0,0 +1,145 @@
...
+            # currently QForte supports up to two-qubit gates
+            count_target = gate_count[gate.target()]
+            count_control = gate_count[gate.control()]
+            # if the target and control wires contain a different
+            # number of gates, we need to add "\qw" to equalize them
+            if count_target < count_control:
+                for i in range(count_control - count_target):
+                    wires[gate.target()] += " & \qw"
+                    gate_count[gate.target()] += 1
+            if count_target > count_control:
```
JonathonMisiewicz, line 111:
```suggestion
            elif count_target > count_control:
```

src/qforte/utils/circuit_to_latex.py
```diff
@@ -0,0 +1,145 @@
...
+                for i in range(count_target - count_control):
+                    wires[gate.control()] += " & \qw"
+                    gate_count[gate.control()] += 1
+            diff = int(gate.control() - gate.target())
+            if gate.gate_id() not in ["cZ", "aCNOT", "acX", "CNOT", "cX"]:
+                raise ValueError("The only two-qubit gates that are currently supported are: CNOT, aCNOT, and cZ!")
+            if gate.gate_id() == "cZ":
```
JonathonMisiewicz, line 118: Are you sure this is right? For the cZ gate, the target qubit looks like a control?

src/qforte/utils/circuit_to_latex.py
imagoulas: Good point!

src/qforte/utils/circuit_to_latex.py
imagoulas: I thought about it, but I decided to hardcoded to "enforce" optimum performance. Of course, I agree with you, so might as well make it an optional argument.

src/qforte/utils/circuit_to_latex.py
```diff
@@ -0,0 +1,145 @@
...
+                wires[gate.target()] += " & \\ctrl{}"
+            else:
+                wires[gate.target()] += " & \\targ{}"
+            if gate.gate_id() == 'aCNOT':
```
JonathonMisiewicz, line 122: I think I'm missing something. cZ and aCNOT gates are distinct, but how does a reader separate acX and CNOT and cX from each other?

imagoulas: The logic is as follows. If we are dealing with a cZ gate, a "control" (`\ctrl{}`) symbol is inserted at the target qubit. Otherwise, a ⊕ symbol (`\targ{}`) is placed at the target qubit. This symbol is common to both aCNOT and CNOT. Subsequently, if we are dealing with an aCNOT, an anti-control (`\octrl{...}`) symbol is placed on the "control" qubit. Otherwise a control {`\ctrl{...}`} symbol is inserted. The difference between the `\octrl` and `ctrl` commands is that one creates an open circle and the other a solid one.

JonathonMisiewicz (approved):
Approved. I'd _like_ a simple regression test, but I won't insist on it.

JonathonMisiewicz:
Looks great!

PR #211, opened 2023-06-19
Add new circuit functions

fevangelista:
## Description
Add functions to manipulate circuits from the Python side. These are introduced mostly for efficiency. As part of this PR, the C++ side was cleaned up a bit, and tests were added.

## User Notes
- [x] `Gate` objects that take a parameter, store the parameter value
- [x] Added functions to create a `Computer` on the python side from a circuit
- [x] Fixed a bug with a function to set circuit parameters
- [x] Upgraded the C++ standard to 17 

## Checklist
- [x] Documented source code
- [x] Checked for redundant headers/imports
- [x] Ready to go!


src/qforte/computer.h
```diff
@@ -131,7 +131,11 @@ class Computer {
     ///  [(basis_1, c_1), (basis_2, c_2), ...]
     void set_state(std::vector<std::pair<QubitBasis, double_c>> state);
 
-    void zero_state();
+    /// reset the quantum computer to the null state (all coefficients are 0)
+    void null_state();
+
+    /// reset the quantum computer to the state |0
```
JonathonMisiewicz, line 137:
```suggestion
    /// reset the quantum computer to the state |0>
```

src/qforte/bindings.cc
```diff
@@ -236,6 +246,22 @@ PYBIND11_MODULE(qforte, m) {
 
     m.def("control_gate", &make_control_gate, "control"_a, "Gate"_a);
 
+    m.def(
+        "prepare_state",
+        [](int nqubit, const std::vector<Gate>& gates) {
+            local_timer t;
```
JonathonMisiewicz, line 252: Does creating this timer accomplish anything?

src/qforte/bindings.cc
```diff
@@ -236,6 +246,22 @@ PYBIND11_MODULE(qforte, m) {
...
+            auto computer = Computer(nqubit);
+            auto circuit = Circuit();
+            for (auto gate : gates) {
```
JonathonMisiewicz, line 255: Should this be a ref?

JonathonMisiewicz (commented):
I'm mostly okay with this, but I'd appreciate test cases.

JonathonMisiewicz:
Great! That addresses all of my recommended changes to the functions.

I'd appreciate the test cases, too.

JonathonMisiewicz:
Please write test cases.

JonathonMisiewicz:
I'm happy with this PR. I leave it to you whether this warrants review by Ilias as well. Adding parameters to gates makes the PR much more ambitious.

JonathonMisiewicz:
For sake of repairing tests, I'm going to request that #226 come in before this PR.

src/qforte/circuit.cc
```diff
@@ -7,19 +7,51 @@
...
         }
     }
+    if (param_idx != param_size) {
+        throw std::runtime_error("Circuit::set_parameters: number of parameters does not match "
+                                 "number of gates with parameters");
+    }
+}
+
+void Circuit::insert_gate(size_t pos, const Gate& gate) {
+    if (pos > gates_.size()) {
```
JonathonMisiewicz, line 30: Why don't we allow `pos = gates_.size`? Isn't that a perfectly legitimate (if inefficient) way to append a gate?

src/qforte/bindings.cc
```diff
@@ -20,13 +21,21 @@ namespace py = pybind11;
...
+        .def(py::init<const Circuit&>())
         .def("add", &Circuit::add_gate)
         .def("add", &Circuit::add_circuit)
         .def("add_gate", &Circuit::add_gate)
         .def("add_circuit", &Circuit::add_circuit)
+        .def("insert_gate", &Circuit::insert_gate)
+        .def("remove_gate", &Circuit::remove_gate)
+        .def("swap_gates", &Circuit::swap_gates)
+        .def("insert_circuit", &Circuit::insert_circuit)
+        .def("remove_gates", &Circuit::remove_gates)
```
JonathonMisiewicz, line 35: Please either include a docstring specifying that this removes gates in a _range_ of positions or rename the function to `remove_gate_interval` or some such. The name makes sense if I already know what the function does, but what the function does isn't obvious if all I know is the name and the function signature.

src/qforte/circuit.h
```diff
@@ -16,15 +16,35 @@ class Circuit {
...
+    /// swap two gates in the circuit
+    void swap_gates(size_t pos1, size_t pos2);
+
+    /// append a circuit at the end of this circuit
     void add_circuit(const Circuit& circ);
 
+    /// insert a circuit at a given position in the circuit
+    void insert_circuit(size_t pos, const Circuit& circ);
+
+    /// remove gates with index i in [pos1, pos2)
```
JonathonMisiewicz, line 42:
```suggestion
    /// remove gates in interval [pos1, pos2)
```

JonathonMisiewicz (commented):
Additional comments. I'll go through another review pass, focusing on test coverage, once these changes are made.

src/qforte/circuit.cc
```diff
@@ -7,19 +7,51 @@
 #include "computer.h"
 #include "sparse_tensor.h"
 
-void Circuit::set_parameters(const std::vector<double>& params) {
-    // need a loop over only gates in state preparation circuit that
-    // have a parameter dependance (if gate_id == Rx, Ry, or Rz)
-    // TODO: make a indexing funciton using a map (Nick)
+void Circuit::set_parameters(const std::vector<std::complex<double>>& params) {
```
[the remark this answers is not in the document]
fevangelista: Good point. I now switched to real parameters. The python bindings accept complex numbers but there is a test to ensure that the imaginary part is zero. This was needed becasue on the py-side we pass complex parameters.

src/qforte/circuit.cc
fevangelista: It's legitimate. I think the code already works when `pos = gates_.size` because we only skip this step if `pos > gates_.size`.

src/qforte/bindings.cc
fevangelista: Thanks for this suggestion.

JonathonMisiewicz (commented):
Thanks! The tests that are here look good, though I do request a couple more.

1. The new Circuit and Computer copy constructors aren't tested. Might I suggest using the copy constructors, doing an equality check, mutating one of the objects, then checking that equality fails?
2. `prepare_computer_from_gate` is not tested.

Now that we _have_ parameters, I think it would be a good follow-up PR to modify the `str` function to include this information.

PR #226, opened 2023-09-08
Fix QPE after SciPy API Change

JonathonMisiewicz:
## Description
SciPy changed the API of their `stats.mode` function but didn't document it. This function adjusts our use of `stats.mode` so as not to raise an error. I have no idea what version first contained the change.

## Checklist
- [x] QPE tests pass again
- [x] Ready to go!

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ conda activate qforte_env
 #### install required packages:
 ```bash
 conda install psi4 -c psi4
-conda install scipy
+conda install scipy>=1.11
 ```
 
 Installation (For Development)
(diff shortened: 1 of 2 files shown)


hrgrimsl (approved):
I see no problems.

imagoulas (commented):
Unfortunately, this change causes the QPE tests to fail on my installation. My scipy version is:
scipy                     1.7.3           py310hfa59a62_0

Would it be too much to make the code agnostic to the version of scipy? For example, we could have an `if` statement checking the version of scipy or use `try` and `except`.
Alternatively, in the installation instructions we can specify:
`conda install scipy>=x.xx`
to avoid such issues.

JonathonMisiewicz:
> Unfortunately, this change causes the QPE tests to fail on my installation. My scipy version is: scipy 1.7.3 py310hfa59a62_0
> 
> Would it be too much to make the code agnostic to the version of scipy? For example, we could have an `if` statement checking the version of scipy or use `try` and `except`. Alternatively, in the installation instructions we can specify: `conda install scipy>=x.xx` to avoid such issues.

So I did some playing with scipy versions. The bug doesn't exist in 1.10.1 but does in 1.11.2. Based on commit history, I suspect [this PR](https://github.com/scipy/scipy/pull/17561) caused the behavior change.

I'd prefer putting a minimum SciPy version in.

PR #231, opened 2023-09-16
Add continuous integration

fevangelista:
## Description
This PR adds continuous integration using GitHub actions. The symmetry and moment corrections tests were disabled because they fail. I would suggest pushing this PR and then dealing with those separately. Let's merge this before PR #211 . Since #209 attempts to do the same, I closed it.

## User Notes
- [x] Tests were disabled.

## Checklist
- [x] Ready to go!


JonathonMisiewicz:
What's the ambiguity in `test_symmetry_attributes`?

I remember that Ilias analyzed the ambiguity in `test_N2_uccsd_pqe`, but I don't think that was written anywhere.

imagoulas:
The ambiguity in `test_symmetry_attributes` is of the same nature as the one encountered in `test_N2_uccsd_pqe`. We did discuss it briefly a few months back.

In both cases, we are dealing with systems of high symmetry (N2 and linear H2O) that involve degenerate orbitals. In QForte, the spin-orbitals are ordered by energy. However, there is an ambiguity in ordering the degenerate orbitals. The `test_symmetry_attributes` test may fail if the degenerate orbitals are ordered differently from what the test anticipates.

Furthermore, the ordering of operators in an SQOpPool depends on the ordering of the orbitals. In addition, the ordering of operators in a fixed-ansatz algorithm is identical to that of the corresponding pool. As a result, the ambiguity of ordering degenerate orbitals creates problems for fixed-ansatze algorithms since, in general, the operators in the pool do not commute. This examples the possible failure of `test_N2_uccsd_pqe`.

fevangelista:
We should talk about how to fix the test cases later. Any issues you see with the current PR?

imagoulas (approved):
Everything looks good to me except for the python-package-conda.yml file, which I am not familiar with its structure. After seeking council form the "Wise One", the file seems OK. I am just curious why the checks with flake8 were removed. In any case, I approve the PR.

fevangelista:
Flake8 was giving problems. I did not investigate that further. It's a feature we can re-introduce later.

PR #245, opened 2024-05-09
Simplification of T, S, Z, and R phase gates added.

imagoulas:
## Description
The T, S, and Z gates are specific instances of the R phase gate. The simplification of pairs of these gates has been incorporated in the simplify function of the Circuit class.

To avoid numerical instabilities associated with large angles, the angle of parametrized gates has been restricted in the -period/2 to period/2 interval, where period = 2pi for the R, cR, and A gates and period = 4pi for the Rx, Ry, Rz, and cRz gates.

To facilitate the simplification of phase gates, the adjoints of T and S now return the corresponding R gate instead of adj(T) and adj(S).

## User Notes
- [X ] Features added

## Checklist
- [x ] Added/updated tests of new features
- [x ] Documented source code
- [ ] Ready to go!

### DIFF
--- a/src/qforte/circuit.cc
+++ b/src/qforte/circuit.cc
@@ -226,6 +226,32 @@ bool Circuit::is_pauli() const {
     return true;
 }
 
+double Circuit::get_phase_gate_parameter(const Gate& gate) {
+    std::unordered_map<GateType, double> gate_parameters = {
+        {GateType::T, M_PI / 4},
+        {GateType::S, M_PI / 2},
+        {GateType::Z, M_PI}
+    };
+
+    if (gate.has_parameter()) {
+        return gate.parameter().value();
+    } else {
+        auto it = gate_parameters.find(gate.gate_type());
+        if (it != gate_parameters.end()) {
+            return it->second;
+        } else {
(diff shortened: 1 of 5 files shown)


src/qforte/circuit.cc
```diff
@@ -226,6 +226,32 @@ bool Circuit::is_pauli() const {
...
+        auto it = gate_parameters.find(gate.gate_type());
+        if (it != gate_parameters.end()) {
+            return it->second;
+        } else {
+            throw std::invalid_argument("Unknown single-qubit phase gate encountered: " + gate.gate_id());
+        }
+    }
+}
+
+double Circuit::simplify_phase_1qubit_gates(const Gate& gate1, const Gate& gate2) {
```
muhan-zhang, line 248: Not wrong. However, the idea to create a method should also be general, meaning that this method should be called for various purposes. I don't see the necessity to have a function here which is just a double addition. You can add the params directly and leave a comment there for clarification.

muhan-zhang (changes requested):
I recommend drop redundant function `simplify_phase_1qubit_gates` in `Circuit` class. I leave a comment in the codes for details.

muhan-zhang (approved):

PR #251, opened 2024-05-16
`get_pauli_permuted_idx` now returns `size_t`

JonathonMisiewicz:
## Description
The `get_pauli_permuted_idx` utility function now returns a `size_t` rather than an `int` as its first argument. The method's only use is as an argument to `std::next`, which logically should be expecting a `size_t`.

Comments, `auto`, and changing other things to `size_t` to match also included.

## Checklist
- [x] Ready to go!

### DIFF
--- a/src/qforte/computer.cc
+++ b/src/qforte/computer.cc
@@ -874,7 +874,7 @@ std::complex<double> Computer::direct_pauli_circ_exp_val(const Circuit& qc) {
     std::vector<int> y_idxs;
     std::vector<int> z_idxs;
 
-    for (const Gate& gate : qc.gates()) {
+    for (const auto& gate : qc.gates()) {
         if (gate.target() < min_qb_idx) {
             min_qb_idx = gate.target();
         }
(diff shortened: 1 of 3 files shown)


src/qforte/computer.cc
```diff
@@ -874,7 +874,7 @@ std::complex<double> Computer::direct_pauli_circ_exp_val(const Circuit& qc) {
     std::vector<int> y_idxs;
     std::vector<int> z_idxs;
 
-    for (const Gate& gate : qc.gates()) {
+    for (const auto& gate : qc.gates()) {
```
imagoulas, line 877: Out of curiosity,  why was this changed to auto?

imagoulas (approved):
Everything looks good to me. Feel free to merge when you are ready.

src/qforte/computer.cc
JonathonMisiewicz: [Using C++ auto is best practice by default.](https://herbsutter.com/2013/08/12/gotw-94-solution-aaa-style-almost-always-auto/)

PR #255, opened 2024-05-20
SRQK uses Circuit copy

JonathonMisiewicz:
## Description
This PR is both simple cleanup and a fix to a subtle bug. If the user passes in `_Uprep`, the old code would ignore that in the slow version but not in the fast version of SRQK. So rather than let the user start from a circuit, they _must_ start from HF, before this PR.

## User Notes
- [x] Fixed a bug where SRQK gives different results in fast/slow mode, if the user passes in `_Uprep`.

## Checklist
- [x] Do you have any idea how to test this one?
- [x] Ready to go!

### DIFF
--- a/src/qforte/abc/algorithm.py
+++ b/src/qforte/abc/algorithm.py
@@ -73,7 +73,7 @@ class Algorithm(ABC):
         self._state_prep_type = state_prep_type
 
         if self._state_prep_type == "occupation_list":
-            if reference == None:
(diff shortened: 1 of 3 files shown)


imagoulas (approved):
Perhaps you can add a test that ensures the consistency of the results between the slow and fast variants of SRQK when `_Uprep` is passed.

Feel free to merge when you are ready,

PR #256, opened 2024-05-20
`QubitOpPool::fill_pool` now takes size_t

JonathonMisiewicz:
## Description
`QubitOpPool::fill_pool` is a qubit basis method, so it shouldn't need to know about any electronic structure, such as references. It just needs to know how many qubits we have.

## User Notes
- [x] `QubitOpPool::fill_pool` now takes the number of qubits, rather than an occupation string.

## Checklist
- [x] Ready to go!


imagoulas (approved):
Good to go!

PR #258, opened 2024-05-21
Clean detection of occupied orbital for PQE

JonathonMisiewicz:
## Description
Removes another todo off the list. The old version made the bad assumption that `_ref` stored information about spatial orbitals rather than spinorbitals, so it wasn't even correct.

## Checklist
- [x] Ready to go!


JonathonMisiewicz:
I need to double-check one detail before this is ready to review...

JonathonMisiewicz:
Okay, yes, the PR is in good shape. It turns out I fixed another silent bug in the code.

imagoulas (approved):

PR #259, opened 2024-05-21
Separate `refprep` and `Uprep`

JonathonMisiewicz:
## Description
Several PQE-like functions need to know how to prepare the reference determinant and assume it can be done by applying `self._Uprep`. This can break down if a user supplied an initial `Uprep` circuit, so this task is now delegated to `refprep`.

## User Notes
- [x] Fixed bugs that could occur if a user supplied an initial circuit for a PQE computation or a calculation with moment corrections

## Checklist
- [x] Ready to go!

### DIFF
--- a/src/qforte/utils/state_prep.py
+++ b/src/qforte/utils/state_prep.py
@@ -4,8 +4,8 @@ import numpy as np
 
 def build_refprep(ref):
     refprep = qforte.Circuit()
-    for j in range(len(ref)):
-        if ref[j] == 1:
+    for j, occupied in enumerate(ref):
+        if occupied:
             refprep.add(qforte.gate("X", j, j))
 
     return refprep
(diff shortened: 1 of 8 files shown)


src/qforte/utils/state_prep.py
```diff
@@ -2,16 +2,13 @@
...
-    if state_prep_type == "occupation_list":
-        for j in range(len(ref)):
-            if ref[j] == 1:
-                Uprep.add(qforte.gate("X", j, j))
-    else:
-        raise ValueError("Only 'occupation_list' supported as state preparation type")
+def build_refprep(ref):
+    refprep = qforte.Circuit()
+    for j in range(len(ref)):
+        if ref[j] == 1:
```
imagoulas, line 8: Perhaps the following is more efficient:

`for j, value in enumerate(ref):`
`    if value:`

imagoulas (approved):
Just one minor suggestion. Feel free to merge when you feel like it.

PR #261, opened 2024-05-22
Added sparse matrix representation of 2qubit gates and updated tests

imagoulas:
### Summary
- Added the sparse matrix representation of two-qubit gates.
- Enhanced QForte to convert all currently implemented gates to sparse matrices.

### Testing
- Comprehensive tests for the new functionality have been added.
- All tests successfully pass.

## User Notes
- [x] Features added

## Checklist
- [x] Documented source code
- [x] Ready to go!

### DIFF
--- a/src/qforte/gate.cc
+++ b/src/qforte/gate.cc
@@ -27,25 +27,67 @@ size_t Gate::control() const { return control_; }
 const complex_4_4_mat& Gate::matrix() const { return gate_; }
 
 const SparseMatrix Gate::sparse_matrix(size_t nqubit) const {
-    size_t nbasis = std::pow(2, nqubit);
-    if (target_ != control_) {
-        throw std::runtime_error("Gate must be a Pauli to convert to matrix!");
-    } else if (target_ >= nqubit) {
+    if (target_ >= nqubit) {
         throw std::runtime_error("Target index is too large for specified nqbits!");
     }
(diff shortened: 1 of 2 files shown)


src/qforte/gate.cc
```diff
@@ -27,25 +27,67 @@ size_t Gate::control() const { return control_; }
...
+
+                        // Combine the target and control bits into a 2-bit state
+                        size_t combined_state = (control_bit << 1) | target_bit;
+
+                        // Check if the combined state matches the gate matrix column index
+                        if (combined_state == j) {
+                            QubitBasis basis_J = basis_I;
+
+                            // Extract the new states for control and target bits
+                            size_t new_control_bit = (i >> 1) & 1;
```
JonathonMisiewicz, line 82: How can the control qubit change? Because it's the control qubit, shouldn't we be able to assume that matrix elements that switch the control qubit are all zero?

I'm _assuming_ that this function is designed to handle any two-qubit gate, and the language of control/target is just a convenience for the most common case.

imagoulas: All of your statements are correct. Currently, all but two two-qubit gates implemented in QForte are controlled. The exceptions are the number-preserving `A` gate and the `SWAP` gate. The target/control designation is used to reflect the most common case.

JonathonMisiewicz (approved):

PR #262, opened 2024-05-22
Remove printing that's specific to gradient-based VQE

JonathonMisiewicz:
## Description
Previously, UCCVQE would print out gradient diagnostics even for computations without a gradient. This is silly, so I removed them.

To clean up the printing, I also tried to unify the to-screen printing and to-file printing. For some reason I don't understand, to-file printing had more digits but not more space for them. There were also some inconsistencies with the amount of whitespace between the two versions. In both cases, I changed them to be more like the to-screen version. The initial `#` for to-file printing is retained.

## User Notes
- [x] There have been minor changes to output printing for UCCVQE.

## Checklist
- [x] Ready to go!

### DIFF
--- a/src/qforte/abc/uccvqeabc.py
+++ b/src/qforte/abc/uccvqeabc.py
@@ -313,9 +313,11 @@ class UCCVQE(UCC, VQE):
 
         if self._k_counter == 1:
             header = "\n    k iteration         Energy               dE"
-            if self._use_analytic_grad: header += "           Ngvec ev      Ngm ev*         ||g||"
+            if self._use_analytic_grad:
+                header += "           Ngvec ev      Ngm ev*         ||g||"
             header += "\n------------------------------------------------------"
-            if self._use_analytic_grad: header += "--------------------------------------------"
+            if self._use_analytic_grad:
+                header += "--------------------------------------------"
             print(header)
             if self._print_summary_file:
                 header.replace("\n ", "\n#  ").replace("\n-", "\n#--")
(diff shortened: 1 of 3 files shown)


src/qforte/abc/uccvqeabc.py
```diff
@@ -312,34 +312,29 @@ def report_iteration(self, x):
...
 
         if self._k_counter == 1:
-            print(
-                "\n    k iteration         Energy               dE           Ngvec ev      Ngm ev*         ||g||"
-            )
-            print(
-                "--------------------------------------------------------------------------------------------------"
-            )
+            header = "\n    k iteration         Energy               dE"
+            if self._use_analytic_grad:
```
imagoulas, line 316: I have a suggestion for future consideration. The decision to print gradient diagnostics should ideally depend on the type of optimizer being used, whether it is gradient-based or gradient-free. Even when numerical gradients are employed, gradient diagnostics should still be printed. However, `scipy.optimize.minimize` does not directly expose numerical gradients to the callback function. Therefore, the current decision to report gradient diagnostics for numerical gradients is based on `self._use_analytic_grad`.

imagoulas, line 316: Looking at `uccnvqe.py`, it appears that a user could potentially request `optimizer = "Nelder-Mead"` and `use_analytic_grad = True`. More or less the same applies to `adaptvqe.py`. To ensure accurate printing of diagnostics, I suggest internally setting `self._use_analytic_grad` to `False` if a gradient-free optimizer is selected. Additionally, a warning message could be printed to inform the user of this change.

imagoulas (approved):
I have only one suggestion to improve the accuracy of the printing. Other than that, everything looks good to me.

imagoulas (approved):

